PR #13, opened 2024-05-01
"Authenticate with OAuth" sample

PaulAllanSturm:


lib/src/samples/authenticate_with_oauth.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:flutter/material.dart';
```
jenmerritt, line 1: Need to add the apache licenses for both sample files

lib/src/samples/authenticate_with_oauth_sample.dart
```diff
@@ -0,0 +1,96 @@
...
+      ),
+    );
+  }
+
+  @override
+  void dispose() async {
+    super.dispose();
+
+    ArcGISEnvironment
+        .authenticationManager.arcGISAuthenticationChallengeHandler = null;
```
jenmerritt, line 52: Maybe a comment for this step too? Just to emphasize it as part of the workflow.

jenmerritt (approved):
Looks good! Just needs the copyright statements, and one other optional doc suggestion.

lib/src/samples/authenticate_with_oauth.dart
PaulAllanSturm: added

lib/src/samples/authenticate_with_oauth_sample.dart
PaulAllanSturm: added

jenmerritt (approved):
👍

PR #16, opened 2024-06-05
Refactor app with json metadata as sources

jenmerritt:
This PR refactors the sample app to:

- Have a structure that will enable the public developers site to pull samples data into the samples directory. As an MVP this involved:
   - An individual directory for each sample
   - Containing a README.metadata.json file and the sourecode
As part of this, I also updated any sample names (names only, not implementation) to match the common designs. And I removed one sample which is not part of the designs (serialize feature collection).

I have also made the Sample Viewer app use the metadata:
   - We generate a single json as a source of the metadata files - this means we don't need to add and load each individual json file as an asset - which when we get up to 100s of files would be a large overhead. This is loaded in `main.dart` and fed to the `SampleListView`.
   - Refactor the sample model class to utilize the metadata. We will want to utilize more of this as we build out the viewer app.
   - Refactor the sample viewer to use a Widget and simplifies by removing the need for an AppBar. We manually list these in samples_widget_list.dart, and access from the viewer using a json key when the sample is selected.
   - Added a `SampleDetailPage` widget which displays the above widget and which we will be able to more easily extend when we implement the ability to view the Readme and code tabs (per the full sample viewer design)

Other updates:
   - Made all the SampleState classes private
   - Renamed the main app to SampleViewerApp from MyApp
   - Updated launch.json
   - Updated README


lib/main.dart
```diff
@@ -28,29 +31,58 @@ void main() {
...
+            ? SampleListView(samples: _samples)
+            : const Center(child: Text('Loading samples...')),
       ),
     );
   }
+
+  void loadSamples() async {
+    final jsonString =
+        await rootBundle.loadString('assets/generated_samples_list.json');
+    Map<String, dynamic> sampleData = jsonDecode(jsonString);
```
PaulAllanSturm, line 80: I think `Map<String, dynamic> sampleData` can just be `final sampleData`?

lib/models/samples_widget_list.dart
```diff
@@ -0,0 +1,71 @@
...
+import 'package:arcgis_maps_sdk_flutter_samples/samples/show_device_location_history/show_device_location_history_sample.dart';
+import 'package:arcgis_maps_sdk_flutter_samples/samples/show_magnifier/show_magnifier_sample.dart';
+import 'package:arcgis_maps_sdk_flutter_samples/samples/show_service_area/show_service_area_sample.dart';
+import 'package:arcgis_maps_sdk_flutter_samples/samples/style_point_with_simple_marker_symbol/style_point_with_simple_marker_symbol_sample.dart';
+import 'package:flutter/material.dart';
+
+// A list of all the Widgets for individual Samples.
+// Used by the Sample Viewer App to display the Widget when a sample is selected.
+// The key is the directory name for the sample which is in snake case. E.g. display_map
+const Map<String, Widget> sampleWidgets = {
```
PaulAllanSturm, line 35: I think we prefer `const sampleWidgets = <String, Widget>{` or however that works

lib/models/samples_widget_list.dart
```diff
@@ -0,0 +1,71 @@
...
+  'query_table_statistics': QueryTableStatisticsSample(),
+  'search_with_geocode': SearchWithGeocodeSample(),
+  'select_features_in_feature_layer': SelectFeaturesInFeatureLayerSample(),
+  'set_basemap': SetBasemapSample(),
+  'show_device_location': ShowDeviceLocationSample(),
+  'show_device_location_history': ShowDeviceLocationHistorySample(),
+  'show_magnifier': ShowMagnifierSample(),
+  'show_service_area': ShowServiceAreaSample(),
+  'style_point_with_simple_marker_symbol':
+      StylePointWithSimpleMarkerSymbolSample()
```
PaulAllanSturm, line 70: add trailing comma (for consistency and merge-friendliness)

PaulAllanSturm:
> removing the need for an AppBar

The app bar gave us a "back" button. And if I remember correctly, on a particular sample that used a drawer, the app bar had a drawer pop-em-out button (but that was mostly accidental). Does the new setup cover that?

jenmerritt:
> The app bar gave us a "back" button. And if I remember correctly, on a particular sample that used a drawer, the app bar had a drawer pop-em-out button (but that was mostly accidental). Does the new setup cover that?

The `AppBar` is instead provided by `SampleDetailPage`, which provides the back navigation functionality, and displays the selected sample name.

For the samples themselves, `AppBar` was only being used to display the title. Find Address and Set Basemap have icons outside of the `AppBar` that load drawers - the drawer itself is defined on the Scaffold, so that behavior is the same. 

Based on other SDK viewers, the app bar should be for the sample viewer application itself - e.g. it may eventually have an `i` icon that loads the readme/code views, and then any sample UI itself is a self contained widget. So nothing has changed / we haven't lost anything in that regard, but then in terms of the app bar being essentially a level higher, it leaves us better placed to extend the viewer application.

PaulAllanSturm:
The AppBar on the "Find Address" sample gives us this weird extra bonus drawer button. If we can get rid of that, that would be preferred.

Screenshot 2024-06-05 at 8 54 19 AM

scripts/generate_samples_list.dart
```diff
@@ -0,0 +1,63 @@
...
+/// Run from arcgis-maps-sdk-flutter-samples root directory
+/// dart run scripts/generate_samples_list.dart
+void main() {
+  generateSamplesList();
+}
+
+/// Compiles all of the individual sample README.metadata.json files into a single JSON asset to feed into the Sample Viewer application.
+/// The output is in assets/generated_samples_list.json
+void generateSamplesList() {
+  Map<String, dynamic> samples = {};
```
PaulAllanSturm, line 15: I think `final samples = <String, dynamic>{};`

scripts/generate_samples_list.dart
PaulAllanSturm, line 1: There's a package layout convention that says this should go in the "tool" directory. https://dart.dev/tools/pub/package-layout#internal-tools-and-scripts

scripts/generate_samples_list.dart
```diff
@@ -0,0 +1,63 @@
...
+  final currentDirectory = Directory.current;
+  final samplesDirectory = Directory('${currentDirectory.path}/lib/samples');
+
+  if (samplesDirectory.existsSync()) {
+    // get all the individual sample directories
+    final individualSampleDirectories = samplesDirectory.listSync();
+
+    // iterate through each directory and get the README.metadata.json file for each sample
+    for (final directory in individualSampleDirectories) {
+      final directoryName = directory.path.split('/').last;
```
PaulAllanSturm, line 29: Prolly should use `Platform.pathSeparator` instead of `/`

scripts/generate_samples_list.dart
```diff
@@ -0,0 +1,63 @@
...
+
+    // write the list of sorted samples to the generated_samples_list.json file
+    final assetsDirectory = Directory('${currentDirectory.path}/assets');
+    final samplesJsonFile =
+        File('${assetsDirectory.path}/generated_samples_list.json');
+    if (samplesJsonFile.existsSync()) {
+      samplesJsonFile.writeAsStringSync(jsonEncode(sortedSamples));
+    }
+  } else {
+    print('Unable to locate the samples directory.');
```
PaulAllanSturm, line 61: Maybe throw this as an exception? To make it clear that it is an error condition. Plus you could reverse the `if` condition, throw the exception, and then the rest of the logic doesn't have to be indented in an `if` block.

scripts/generate_samples_list.dart
```diff
@@ -0,0 +1,63 @@
...
+    final individualSampleDirectories = samplesDirectory.listSync();
+
+    // iterate through each directory and get the README.metadata.json file for each sample
+    for (final directory in individualSampleDirectories) {
+      final directoryName = directory.path.split('/').last;
+      final metadataFile = (directory as Directory)
+          .listSync()
+          .firstWhere((file) => file.path.contains('metadata.json')) as File;
+      // get the json string for the sample
+      Map<String, dynamic> sampleJsonContent =
```
PaulAllanSturm, line 34: `final sampleJsonContent`

scripts/generate_samples_list.dart
```diff
@@ -0,0 +1,63 @@
...
+    for (var key in sortedSamples.keys) {
+      print(key);
+    }
+    print('Total samples: ${sortedSamples.keys.length}');
+
+    // write the list of sorted samples to the generated_samples_list.json file
+    final assetsDirectory = Directory('${currentDirectory.path}/assets');
+    final samplesJsonFile =
+        File('${assetsDirectory.path}/generated_samples_list.json');
+    if (samplesJsonFile.existsSync()) {
```
PaulAllanSturm, line 57: We should make this work whether or not `samplesJsonFile` already exists, since we're going to be blasting all new content on it in either case.

jenmerritt:
> The AppBar on the "Find Address" sample gives us this weird extra bonus drawer button. If we can get rid of that, that would be preferred. ("Set Basemap" has it too.)

Yeah that's gone with the new architecture, so for Find Address there's just the icon next to the search bar. Set Basemap just has the basemap icon, bottom right.
I think Flutter automatically adds the hamburger if a drawer is created (either as the leading or trailing, depending where the drawer is configured start/end) and you have to overwrite in the AppBar properties. But since the AppBar is detached from the sample now it doesn't appear anymore anyway.

lib/main.dart
jenmerritt: So if I don't define `Map<String, dynamic>` it's just type `dynamic` being returned. Which works, but I guess I naturally feel more comfortable defining the type there. What do you think?

lib/main.dart
```diff
@@ -28,29 +31,58 @@ void main() {
...
+    for (final s in sampleData.entries) {
+      _samples.add(Sample.fromJson(s.value));
+    }
+    setState(() {
+      _ready = true;
+    });
```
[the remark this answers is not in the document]
jenmerritt: Yep, code completion doesn't help with remembering this one. Fixed.

scripts/generate_samples_list.dart
jenmerritt: My thinking here with these checks was more if say the script was run from the wrong directory (`Directory.current`), because then `samplesDirectory` could be the wrong path. I'll just throw here too if it can't find it per your other comment.

scripts/generate_samples_list.dart
jenmerritt: See what you think for this and the other `exists` check

scripts/generate_samples_list.dart
PaulAllanSturm: How about if the check was "does the assets directory exist"? If there's an assets directory, we can safely assume that we are supposed to fill it with a `generated_samples_list.json` file, whether or not there's a `generated_samples_list.json` file already there. My concern is that I can easily see my future self deleting the generated file and getting surprised that the script no longer works.

scripts/generate_samples_list.dart
jenmerritt: Will leave per the above comment - since jsonDecode returns `dynamic` 👍

PaulAllanSturm: nevermind about this one, per other discussion

scripts/generate_samples_list.dart
PaulAllanSturm: Right yeah, that kind of exception. But I would do both checks up top, so they act as preconditions -- if there's something amiss about the setup, bail up front before doing any work. So like, it would check for both `lib/samples/` and `assets/` at the current directory. If both appear, then we're satisfied that we're good to go. If either is missing, throw up our hands and say we're not dealing with this.

tool/generate_samples_list.dart
```diff
@@ -0,0 +1,74 @@
...
+void generateSamplesList() {
+  final pathSeparator = Platform.pathSeparator;
+  final samples = <String, dynamic>{};
+
+  print('Generating samples list....');
+
+  // check for the required directories
+  final currentDirectory = Directory.current;
+  final samplesDirectory = Directory(
+      '${currentDirectory.path}${pathSeparator}lib${pathSeparator}samples');
```
PaulAllanSturm, line 23: I wish there was a cleaner way to assemble platform-aware paths, but there just isn't. And actually here this might be overkill, because `Directory` is smart enough to understand `/` on both platforms. Unfortunately there's a minefield -- `Directory` and `File` are smart enough, but `Link` and `split()` aren't smart enough. Being explicit about the `pathSeparator` is definitely safest.

In another PR I used `Directory(Uri(pathSegments: ['..', '..', thing1, thing2]).toFilePath())` but I dunno if that's any better. Another option is something like `Directory([currentDirectory.path, 'libs', 'samples'].join(pathSeparator))`. I can't decide if we should standardize on which approach.

PaulAllanSturm (approved):

tool/generate_samples_list.dart
jenmerritt: Let's add it to our coding standards issue (I'll do that), and that way others can weigh in on any preferences.

PR #17, opened 2024-06-05
Use runner for individual samples

jenmerritt:
As part of the refactor I considered this change, but wanted to keep it separate incase others prefer the existing behavior. It removes the secondary sample files and instead creates a runner that displays a sample.

For me, this reduces overhead by not having to have 2 sourcecode files for every sample and keeps things cleaner. There is no AppBar/title but I feel that's not necessary anyway.

Going forward, we could also feasibly add a run config for each sample to `launch.json` making it even easier for users to run a particular sample, as they'd just choose it from the dropdown in VS Code. Though this will be easier/quicker to automate with a script so I've left that out for now and just updated the "run custom sample" config.


jenmerritt:
Agreed - I've created an internal issue for automating more in the samples 👍

kmueller-gis (approved):
Reducing the number of files to add and edit is a great thing. This change makes a lot of sense since the files getting removed were nearly identical for every sample.

jenmerritt:
Goodness knows what's happened with these conflicts. I'm closing and cherry picked these changes to a clean branch to save time.

PR #25, opened 2024-06-11
Shokhina toj/1537 update reverse geocode design

shokhinaToj:
Current Flutter sample: https://github.com/Esri/arcgis-maps-sdk-flutter-samples/tree/v.next/lib/samples/find_address_with_reverse_geocode 

- Removed search bar, drawer, and San Diego locatorTask from this sample, following the common design
- Set an initial view point based on the sample implementation code
- Used reverseGeocode() function that returns Future and printing out the full address on the AlertDialog()
- Due to the absense of Callout function, I used AlertDialog() to display the address based on other samples implemented previously (Ex. search with geocode sample)
- Updated the README metadata file based on the common design README


lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -47,158 +51,78 @@ class _FindAddressWithReverseGeocodeSampleState
...
-
-    _geocodeParameters.minScore = 75;
-    _geocodeParameters.resultAttributeNames
-        .addAll(['Place_addr', 'Match_addr']);
+    _mapViewController.setViewpoint(_initialViewpoint);
 
     await Future.wait([
       _worldLocatorTask.load(),
-      _sanDiegoLocatorTask.load(),
     ]);
```
PaulAllanSturm, line 84: The `Future.wait([])` thing was just because there were two Futures to wait on. Since we're down to one, we don't need `Future.wait([])` any more -- it can just be `await _worldLocatorTask.load();`

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -47,158 +51,78 @@ class _FindAddressWithReverseGeocodeSampleState
...
+      final address = firstResult.attributes['LongLabel'] as String;
       showDialog(
         context: context,
         builder: (BuildContext context) {
-          return AlertDialog(content: Text('$matchAddr\n($x, $y)'));
+          return AlertDialog(content: Text(address));
         },
       );
+    } on StateError {
+      return;
```
PaulAllanSturm, line 118: Is this `StateError` part taking the place of the `if (mounted)` check? We do `if (mounted)` in a few other places -- we should decide which is the best approach.

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -47,158 +51,78 @@ class _FindAddressWithReverseGeocodeSampleState
...
+    reverseGeocodeParameters.maxResults = 1;
+    final reverseGeocodeResult = await _locatorTask.reverseGeocode(
+      location: normalizedTapPoint as ArcGISPoint,
+      parameters: reverseGeocodeParameters,
+    );
+    if (reverseGeocodeResult.isEmpty) return;
+
+    try {
+      final firstResult = reverseGeocodeResult.first;
+      final address = firstResult.attributes['LongLabel'] as String;
```
PaulAllanSturm, line 110: Why is this line inside the `try` block? Is it possible to get a `StateError` out of it? If not, I would think it should go before the `try` block.

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -47,158 +51,78 @@ class _FindAddressWithReverseGeocodeSampleState
...
+    if (mapTapPoint == null) return;
 
-    if (mounted) {
-      final graphic = identifyGraphicsOverlayResult.graphics.first;
-      final center = graphic.geometry!.extent.center;
-      final x = center.x.toStringAsFixed(3);
-      final y = center.y.toStringAsFixed(3);
-      final matchAddr = graphic.attributes['Match_addr'];
+    final normalizedTapPoint =
+        GeometryEngine.normalizeCentralMeridian(geometry: mapTapPoint);
```
PaulAllanSturm, line 97: I see that `normalizeCentralMeridian` in the reference sample code, but I'm not sure it's necessary. I think it might just be adding extra complication. @jenmerritt  do you know if we need to be doing this?

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
shokhinaToj: I am not sure if it was StateError, but this line was throwing some error when I indicated a wrong attribute key. But since I know "LongLabel" is the right key, it shouldn't be throwing any errors - I will take these two lines out!

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
shokhinaToj: I just checked -- StateError is a more general error that catches things related to object states however "if(mounted)" will be checking if the state is still mounted after async reverseGeocdo function is ran so it is safer to keep "if(mounted)" I think. Jen is also using it in many other places to display AlertDialog.

lib/samples/find_address_with_reverse_geocode/README.metadata.json
```diff
@@ -3,9 +3,9 @@
     "description": "Use an online service to find the address for a tapped point.",
     "ignore": false,
     "images": [],
-    "keywords": [],
+    "keywords": ["address", "geocode", "locate", "reverse geocode", "search"],
     "redirect_from": [],
-    "relevant_apis": [],
+    "relevant_apis": ["GeocodeParameters", "LocatorTask", "ReverseGeocodeParameters"],
     "snippets": [],
```
shokhinaToj, line 9: @jenmerritt Not sure if these updates are right, I tried to follow the README from common design

shokhinaToj, line 9: @jenmerritt I followed the common design README to populate this metadata but need to confirm if everything is correct

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
jenmerritt: Because we have wraparoundmode enabled we should normalize before passing to the geocode operation as if you tap a point across the date line it messes things up. We could remove this and set `WrapAroundMode.disabled` on the mapview instead. But given other implementations are using this I think we should follow suit.

One thing we need to do in the sample (which I'll go through with you on our call @shokhinaToj ) is to add comments in the code, so we can add a little note on this too.

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -47,158 +51,74 @@ class _FindAddressWithReverseGeocodeSampleState
...
       key: _scaffoldStateKey,
-      endDrawer: Drawer(
-        child: SafeArea(
-          child: Padding(
-            padding: const EdgeInsets.all(10),
-            child: buildServiceControls(context),
-          ),
-        ),
-      ),
       resizeToAvoidBottomInset: false,
```
jenmerritt, line 54:
```suggestion
```
I think we can remove this now too, as it was used to position the keyboard when we had search functionality.

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -47,158 +51,74 @@ class _FindAddressWithReverseGeocodeSampleState
...
-
   void onMapViewReady() async {
-    final map = ArcGISMap.withBasemapStyle(BasemapStyle.arcGISImagery);
+    final map = ArcGISMap.withBasemapStyle(BasemapStyle.arcGISTopographic);
 
     final image = await ArcGISImage.fromAsset('assets/pin_circle_red.png');
     final pictureMarkerSymbol = PictureMarkerSymbol.withImage(image);
     pictureMarkerSymbol.width = 35;
     pictureMarkerSymbol.height = 35;
     pictureMarkerSymbol.offsetY = pictureMarkerSymbol.height / 2;
```
jenmerritt, line 74:
```suggestion
    final pictureMarkerSymbol = PictureMarkerSymbol.withImage(image)
    ..width = 35
    ..height = 35
    ..offsetY = pictureMarkerSymbol.height / 2;
```

I think we should promote cascade notation in our samples, such as in places like this. What are your thoughts @PaulAllanSturm ? As long as it doesn't muddy the story, I think it's neat when we set something up without a specific constructor.

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -47,158 +51,74 @@ class _FindAddressWithReverseGeocodeSampleState
...
-    if (locatorTask == null) return;
-
-    setState(() => _locatorTask = locatorTask);
-    Future.delayed(
-      const Duration(milliseconds: 250),
-      () => _scaffoldStateKey.currentState!.closeEndDrawer(),
+  void displayTappedPoint(Geometry mapPoint) {
+    final pointGraphic = Graphic(
+      geometry: mapPoint,
     );
```
jenmerritt, line 121:
```suggestion
    final pointGraphic = Graphic(geometry: mapPoint);
```
Since this is just one property we can make it one line

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
PaulAllanSturm: It was new to me, but I think we can presume a Dart developer is familiar with cascade notation. I think it's clean and idiomatic so we should use cascade notation.

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -47,158 +51,74 @@ class _FindAddressWithReverseGeocodeSampleState
...
+        _mapViewController.screenToLocation(screen: localPosition);
+    if (mapTapPoint == null) return;
 
-    if (identifyGraphicsOverlayResult.graphics.isEmpty) return;
+    final normalizedTapPoint =
+        GeometryEngine.normalizeCentralMeridian(geometry: mapTapPoint);
+    displayTappedPoint(normalizedTapPoint!);
+
+    final reverseGeocodeParameters = ReverseGeocodeParameters();
+    reverseGeocodeParameters.maxResults = 1;
```
jenmerritt, line 99:
```suggestion
    final reverseGeocodeParameters = ReverseGeocodeParameters()
    ..maxResults = 1;
```
Another place we can use [cascade notation](https://dart.dev/language/operators#cascade-notation).

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
PaulAllanSturm: Oh, I though WrapAroundMode was disabled by default. Everything I said after that was backwards, don't mind me.

But! I do have one other actually good point: `normalizedTapPoint` might be `null`, so instead of `!` maybe it should do:
```dart
if (normalizedTapPoint == null) return;
```
Seeing a `!` is always a little scary unless we can directly reason that it can't be null. Here we're relying on the internal machinations of `GeometryEngine`, so we shouldn't assume too much.

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
shokhinaToj: Yes I think you are right - this way is more concise and clearer

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -46,159 +49,73 @@ class _FindAddressWithReverseGeocodeSampleState
...
 
     final image = await ArcGISImage.fromAsset('assets/pin_circle_red.png');
-    final pictureMarkerSymbol = PictureMarkerSymbol.withImage(image);
-    pictureMarkerSymbol.width = 35;
-    pictureMarkerSymbol.height = 35;
-    pictureMarkerSymbol.offsetY = pictureMarkerSymbol.height / 2;
+    final pictureMarkerSymbol = PictureMarkerSymbol.withImage(image)
+    ..width = 35
+    ..height = 35
+    ..offsetY = pictureMarkerSymbol.height / 2;
```
shokhinaToj, line 71: @jenmerritt line 71 is giving error because pictureMarkerSymbol is being referenced before it is done being declared - is it okay if reset it to what it was (line 131)

jenmerritt: Ohh yeah, yeah I'd just do that line separately / revert that bit to what you had 👍

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
shokhinaToj: Gotcha! I will make sure to apply this change

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -46,158 +49,84 @@ class _FindAddressWithReverseGeocodeSampleState
...
+    if (reverseGeocodeResult.isEmpty) return;
 
-    if (geocodeResults.isEmpty) return;
+    // get and show the address.
+    final firstResult = reverseGeocodeResult.first;
+    final cityString = firstResult.attributes['City'] ?? "" as String;
+    final addressString = firstResult.attributes['Address'] ?? "" as String;
+    final stateString = firstResult.attributes['RegionAbbr'] ?? "" as String;
+    List<String> resultString = [addressString, cityString, stateString];
+    String combinedString =
```
PaulAllanSturm, line 121: As a coding style, we want to use type inference whenever we can, so these can be `final resultString =` and `final combinedString =`

PaulAllanSturm: and actually `resultString` should be `resultStrings` since it's a list

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -46,158 +49,84 @@ class _FindAddressWithReverseGeocodeSampleState
...
-          ),
-          onPressed: () => _scaffoldStateKey.currentState!.openEndDrawer(),
-        ),
-      ],
-    );
-  }
-
   void onMapViewReady() async {
-    final map = ArcGISMap.withBasemapStyle(BasemapStyle.arcGISImagery);
+    // create an instance of a map with ESRI topographic basemap.
```
PaulAllanSturm, line 66: Oh, we're supposed to be Esri now, not ESRI

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
shokhinaToj: Okay! Is there a specific reason why we use type inference. To reduce human error?

PaulAllanSturm: Basically let the compiler do the work. Less typing. And like, if you later change the type of the thing on the right side of the equation, you don't also have to do the bookkeeping on the left side to keep up -- the type will propagate wherever it needs to go and the compiler will tell you if it runs into a mismatch somewhere.

shokhinaToj: That makes sense, thanks Paul!

PaulAllanSturm (approved):

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
shokhinaToj: ohh that makes sense, thanks for the explanation!

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -46,158 +49,84 @@ class _FindAddressWithReverseGeocodeSampleState
...
+      parameters: reverseGeocodeParameters,
     );
+    if (reverseGeocodeResult.isEmpty) return;
 
-    if (geocodeResults.isEmpty) return;
+    // get and show the address.
+    final firstResult = reverseGeocodeResult.first;
+    final cityString = firstResult.attributes['City'] ?? "" as String;
+    final addressString = firstResult.attributes['Address'] ?? "" as String;
+    final stateString = firstResult.attributes['RegionAbbr'] ?? "" as String;
```
jenmerritt, line 119:
```suggestion
    final cityString = firstResult.attributes['City'] ?? '';
    final addressString = firstResult.attributes['Address'] ?? '';
    final stateString = firstResult.attributes['RegionAbbr'] ?? '';
```
We should use single quotes here and actually while the null check you've added is great, we don't need to cast the second half to a String.

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -46,158 +49,84 @@ class _FindAddressWithReverseGeocodeSampleState
...
+    final resultStrings = [addressString, cityString, stateString];
+    final combinedString =
+        resultStrings.where((str) => str.isNotEmpty).join(", ");
```
jenmerritt, line 122:
```suggestion
        resultStrings.where((str) => str.isNotEmpty).join(', ');
```

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode_sample.dart
```diff
@@ -46,158 +49,84 @@ class _FindAddressWithReverseGeocodeSampleState
...
-          child: Padding(
-            padding: const EdgeInsets.all(10),
-            child: buildServiceControls(context),
-          ),
-        ),
-      ),
-      resizeToAvoidBottomInset: false,
       body: Stack(
         children: [
+          // add a map view.
```
jenmerritt, line 54:
```suggestion
          // add a map view to the widget tree and set a controller.
```

jenmerritt (commented):
Thanks @shokhinaToj - the readme updates look great. I'll ping you directly about some of the analyzer warnings we can look out for in the dart formatter. 
Just a couple of tidy ups in the code and then some tweaks to the comments - we're still establishing patterns here so it will get easier to follow existing samples as we build more out, especially around comments. So appreciate your patience and help working through this process! You've done a great job with the sample UI updates 👍

PR #31, opened 2024-06-18
Apply Class Breaks Renderer To Sublayer Sample

shokhinaToj:
Current Flutter sample: https://github.com/Esri/arcgis-maps-sdk-flutter-samples/tree/v.next/lib/samples/apply_class_breaks_renderer_to_sublayer

Applied common design to apply-class-breaks-renderer-to-sublayer sample.

The sample had missing button for applying the renderer and resetting the map. I added these two buttons, included necessary comments, added the README, and generated the metadata.

The README file was created based on the example of common designs README for this sample but multiple steps and API names were altered/removed - this needs to be reviewed.


jenmerritt:
The workflow of this one isn't quite right, I'll message you directly about this one.

lib/samples/apply_class_breaks_renderer_to_sublayer/apply_class_breaks_renderer_to_sublayer_sample.dart
```diff
@@ -28,42 +28,97 @@ class ApplyClassBreaksRendererToSublayerSample extends StatefulWidget {
...
+  // set the ready state to false
+  bool _ready = false;
 
   @override
   Widget build(BuildContext context) {
     return Scaffold(
-      body: ArcGISMapView(
-        controllerProvider: () => _mapViewController,
-        onMapViewReady: onMapViewReady,
+      body: SafeArea(
```
PaulAllanSturm, line 41: probably `top: false` needed for the SafeArea

lib/samples/apply_class_breaks_renderer_to_sublayer/apply_class_breaks_renderer_to_sublayer_sample.dart
```diff
@@ -28,42 +28,97 @@ class ApplyClassBreaksRendererToSublayerSample extends StatefulWidget {
...
+  }
+
+  void resetRoutes() {
+    // clear the map
+    map.operationalLayers.clear();
+    setState(() => _rendered = false);
   }
 
   ClassBreaksRenderer createPopulationClassBreaksRenderer() {
+    // create coors for the class breaks
```
PaulAllanSturm, line 114: "coors" should be "colors"

lib/samples/apply_class_breaks_renderer_to_sublayer/apply_class_breaks_renderer_to_sublayer_sample.dart
```diff
@@ -108,6 +164,7 @@ class _ApplyClassBreaksRendererToSublayerSampleState
         maxValue: 10110975,
         symbol: classSymbol5);
```
PaulAllanSturm, line 165: I think we would add a trailing comma for all these classBreakX variables

lib/samples/apply_class_breaks_renderer_to_sublayer/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+![Image of apply class breaks renderer to sublayer](apply_class_breaks_renderer_to_sublayer.png)
+
+## Use case
+
+A layer showing animal populations contains sublayers for different species. A renderer could be applied which gives each sublayer a different color, so that populations of each species can be compared visually.
+
+## How to use the sample
+
+Wait for the map image layer to load. Click the 'Change Sublayer Renderer' button to apply a unique value renderer to see different population ranges in the counties sub-layer data.
```
jenmerritt, line 13:
```suggestion
Wait for the map image layer to load. Tap the 'Change Sublayer Renderer' button to apply a unique value renderer to see different population ranges in the counties sub-layer data.
```

lib/samples/apply_class_breaks_renderer_to_sublayer/apply_class_breaks_renderer_to_sublayer_sample.dart
```diff
@@ -28,42 +28,97 @@ class ApplyClassBreaksRendererToSublayerSample extends StatefulWidget {
 class _ApplyClassBreaksRendererToSublayerSampleState
     extends State<ApplyClassBreaksRendererToSublayerSample> {
   final _mapViewController = ArcGISMapView.createController();
+  // create a map with a basemap
```
jenmerritt, line 31:
```suggestion
  // create a map with a basemap style
```

lib/samples/apply_class_breaks_renderer_to_sublayer/apply_class_breaks_renderer_to_sublayer_sample.dart
```diff
@@ -28,42 +28,97 @@ class ApplyClassBreaksRendererToSublayerSample extends StatefulWidget {
...
+        ),
       ),
     );
   }
 
-  void onMapViewReady() async {
-    final map = ArcGISMap.withBasemapStyle(BasemapStyle.arcGISStreets);
-
+  void renderLayer() async {
+    // create an image layer
```
jenmerritt, line 79: We are missing the first step in the sample. So the order should be:

1. Create a map and set the sublayer to the operational layers. This will then display the sublayer with it's default renderer (which is a black outline style).
2. Capture the default renderer (`defaultRenderer = subLayer.renderer`) so it can reapplied when the user resets.
3. When the user taps the "change sublayer renderer" button - the renderer is changed to the class breaks renderer (the blue style).
4. When the user taps "reset", the renderer is set back to the default renderer. (`subLayer.renderer = defaultRenderer`).

So when the sample first loads, the layer should be visible and all that's happening is the renderer is being updated / reset.

lib/samples/apply_class_breaks_renderer_to_sublayer/apply_class_breaks_renderer_to_sublayer_sample.dart
```diff
@@ -28,17 +28,22 @@ class ApplyClassBreaksRendererToSublayerSample extends StatefulWidget {
...
+  // create a map with a basemap style
   final map = ArcGISMap.withBasemapStyle(BasemapStyle.arcGISStreets);
   // set the rendered state to false
   bool _rendered = false;
   // set the ready state to false
   bool _ready = false;
+  // create an image sublayer
+  late ArcGISMapImageSublayer countiesSublayer;
+  // create an renderer
+  late Renderer defaultRenderer;
```
PaulAllanSturm, line 40: I think generally we've been naming all member variables with a leading `_`. So that should be `_countiesSublayer`, `_defaultRenderer`, and also `_map` a couple lines up.

lib/samples/apply_class_breaks_renderer_to_sublayer/apply_class_breaks_renderer_to_sublayer_sample.dart
```diff
@@ -28,42 +28,109 @@ class ApplyClassBreaksRendererToSublayerSample extends StatefulWidget {
...
   final _mapViewController = ArcGISMapView.createController();
+  // create a map with a basemap style
+  final map = ArcGISMap.withBasemapStyle(BasemapStyle.arcGISStreets);
+  // set the rendered state to false
+  bool _rendered = false;
+  // set the ready state to false
+  bool _ready = false;
+  // create an image sublayer
+  late ArcGISMapImageSublayer countiesSublayer;
+  // create an renderer
```
jenmerritt, line 39:
```suggestion
  // create a renderer
```

lib/samples/apply_class_breaks_renderer_to_sublayer/apply_class_breaks_renderer_to_sublayer_sample.dart
```diff
@@ -28,42 +28,109 @@ class ApplyClassBreaksRendererToSublayerSample extends StatefulWidget {
 class _ApplyClassBreaksRendererToSublayerSampleState
     extends State<ApplyClassBreaksRendererToSublayerSample> {
   final _mapViewController = ArcGISMapView.createController();
+  // create a map with a basemap style
+  final map = ArcGISMap.withBasemapStyle(BasemapStyle.arcGISStreets);
+  // set the rendered state to false
+  bool _rendered = false;
+  // set the ready state to false
+  bool _ready = false;
```
jenmerritt, line 36:
```suggestion
  // set the ready state to false
  var _ready = false;
```

lib/samples/apply_class_breaks_renderer_to_sublayer/apply_class_breaks_renderer_to_sublayer_sample.dart
```diff
@@ -28,42 +28,109 @@ class ApplyClassBreaksRendererToSublayerSample extends StatefulWidget {
...
+                  // apply renderer button
+                  ElevatedButton(
+                    onPressed: !_rendered && _ready ? renderLayer : null,
+                    child: const Text('Change Sublayer Renderer'),
+                  ),
+                  // reset button
+                  ElevatedButton(
+                    onPressed: _rendered ? resetRoutes : null,
+                    child: const Text('Reset'),
+                  ),
```
jenmerritt, line 73: Sorry this is probably my bad for my initial comments about having the default renderer and not getting this fully correct then, there's a lot of sample designs flying around! I've just reviewed the design and readme and there's no mention of a reset button so it looks like we just need the "Change Sublayer Renderer" button to toggle which renderer is being used. I think this also means we can get rid of this `_renderered` boolean which is making things feel more complex than they need to be.

We can have a call to go through this sample if this doesn't make sense.

lib/samples/apply_class_breaks_renderer_to_sublayer/apply_class_breaks_renderer_to_sublayer_sample.dart
jenmerritt: See my other comment below, I don't think we need 2 boolean checks.

lib/samples/apply_class_breaks_renderer_to_sublayer/apply_class_breaks_renderer_to_sublayer_sample.dart
shokhinaToj: Alright, no worries! It should be easy to change few things around

jenmerritt: I was thinking it would toggle it but yeah you're right the design doesn't specify that. I guess you will still need the `_renderer` flag then to be able to disable the button once it's renderer.

i.e. you'll still need: `onPressed: _rendered ? changeRenderer : null,`

jenmerritt (approved):

PaulAllanSturm (approved):

PR #35, opened 2024-06-24
Filter by Definition Expression or Display Filter Sample

shokhinaToj:
Current Flutter sample: https://github.com/ArcGIS/arcgis-maps-sdk-flutter-samples/blob/main/lib/src/samples/filter_by_definition_expression_or_display_filter_sample.dart

The sample works as expected but didn't display number of features in the current map view extent. This feature was implemented and needs review(the numbers for each case). README file was created, comments were added, and metadata was generated.


shokhinaToj:
Screenshot of one of the cases (display filters) was added - do we want a screenshot of something else or multiple screenshots for all cases @jenmerritt

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
```diff
@@ -27,23 +27,27 @@ class FilterByDefinitionExpressionOrDisplayFilterSample extends StatefulWidget {
 
 class _FilterByDefinitionExpressionOrDisplayFilterSampleState
     extends State<FilterByDefinitionExpressionOrDisplayFilterSample> {
+  // create a map view controller
   final _mapViewController = ArcGISMapView.createController();
-  final _featureLayer = FeatureLayer.withFeatureTable(
+  final FeatureLayer _featureLayer = FeatureLayer.withFeatureTable(
```
PaulAllanSturm, line 32: Why was the "FeatureLayer" type added here? I think it's unnecessary.

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
```diff
@@ -54,21 +58,26 @@ class _FilterByDefinitionExpressionOrDisplayFilterSampleState
         child: Column(
```
PaulAllanSturm, line 58: Does the SafeArea need a "top: false"? (So that the map goes all the way up under the notch when running with sample_runner.dart)

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
```diff
@@ -82,24 +91,61 @@ class _FilterByDefinitionExpressionOrDisplayFilterSampleState
...
+
+  void calculateFeatureCount() async {
+    // count the number of features
+    final queryParameters = QueryParameters();
+    // get the current extent of the map view
+    queryParameters.geometry = _mapViewController
+        .getCurrentViewpoint(viewpointType: ViewpointType.boundingGeometry)
+        ?.targetGeometry
+        .extent;
+    queryParameters.whereClause = '1=1';
```
PaulAllanSturm, line 135: Is this line necessary? I think it should work without a `whereClause` as long as the other values are set properly on `_featureLayer`.

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
```diff
@@ -82,24 +91,61 @@ class _FilterByDefinitionExpressionOrDisplayFilterSampleState
...
+
+    // query the feature count
+    final featureCount = await _featureLayer.featureTable!
+        .queryFeatureCount(queryParameters: queryParameters);
+
+    // show the feature count in an alert dialog
+    if (mounted) {
+      showDialog(
+        context: context,
+        builder: (BuildContext context) {
```
PaulAllanSturm, line 145: I think we prefer `(context)` instead of `(BuildContext context)` for closures.

jenmerritt:
> Screenshot of one of the cases (display filters) was added - do we want a screenshot of something else or multiple screenshots for all cases

The common design is desktop to hard to match exactly, so I would follow the iOS (old samples) image which has a filter applied with the popup of number of features displayed. This then demonstrates the functionality of the sample - i.e. a filter has been applied and the number of features is significantly reduced.

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
```diff
@@ -82,24 +91,61 @@ class _FilterByDefinitionExpressionOrDisplayFilterSampleState
...
+          return AlertDialog(content: Text('$featureCount features found.'));
```
jenmerritt, line 146: Let's have a Title 'Current feature count' and Content as `$featureCount features`. This aligns with the common design and iOS implementation, and makes the dialog look a little tidier, since we will be keeping the dialog for this sample and not looking to implement a callout or any other UI alternative.

lib/samples/filter_by_definition_expression_or_display_filter/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Alternatively, display filters limit which features are drawn, but retain all features in queries and when processing. Definition queries and display filters can be used together on a layer, but definition queries limit the features available in the layer, while display filters only limit which features are displayed.
+
+In this sample you can filter a dataset of tree quality selecting for only those trees which require maintenance or are damaged.
+
+## How to use the sample
+
+Tap on the apply expression button to limit the features requested from the feature layer to those specified by the SQL query definition expression. This option not only narrows down the results that are drawn, but also removes those features from the layer's attribute table. To filter the results being drawn without modifying the attribute table, tap on the button to apply the filter instead. Tap on the reset button to remove the definition expression or display filter on the feature layer, which returns all the records.
```
jenmerritt, line 17: Nice one picking up these "tap" updates 👍 🙂

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
```diff
@@ -27,23 +27,27 @@ class FilterByDefinitionExpressionOrDisplayFilterSample extends StatefulWidget {
...
-  final _featureLayer = FeatureLayer.withFeatureTable(
+  final FeatureLayer _featureLayer = FeatureLayer.withFeatureTable(
       ServiceFeatureTable.withUri(Uri.parse(
           'https://services2.arcgis.com/ZQgQTuoyBrtmoGdP/arcgis/rest/services/SF_311_Incidents/FeatureServer/0')));
 
   @override
   void initState() {
     super.initState();
-
+    // create a map with a topographic basemap style
```
jenmerritt, line 39: This is a good opportunity to discuss this setup pattern that we talked about last week. We should probably establish a pattern for what we consider best practice in terms of setting up the map so that it doesn't raise questions with those using our samples as a guide. It will also come up when we implement tutorials etc. on the developer website.

In most of the samples we've done the last few weeks we've used `_onMapViewReady` to do configuration. Sometimes that's because we need the map view to actually be ready before we can do "stuff".
It feels to me like it's good to use `initState` if we can - but then should we be using `initState` for everything that can be done up front, and then use `onMapViewReady` for anything that requires the map view to be instantiated? An alternative that avoids separation of setup code and keeps the sample a little more linear could be to say that if we need `onMapViewReady` then set everything up there, as we have been doing, and if not use `initState`, like this sample.

@PaulAllanSturm I'd be interested in your thoughts here.

jenmerritt: > Which implies that we abandon initState(). It's not bad or wrong, but only onMapViewReady() allows us to be consistent across samples for all use cases.

Sounds good. And I completely agree that samples often need compromise to be readable vs hyper-optimized.

So @shokhinaToj , for this sample, could you move the code in `initState` into `onMapViewReady`? And if you want to go through anything in the above discussion, just give me a shout.

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
shokhinaToj: Yes you are right - I think before I wasn't initializing it here and type was needed but now it is unnecessary!

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
PaulAllanSturm: I suppose it should say, "..., if possible". Definitely you can't avoid `late` in all cases.

One situation I'm thinking of is something like this:
```dart
final _graphicsOverlay = GraphicsOverlay()
  ..renderer = SimpleRenderer(symbol: defaultMarkerSymbol)
  ..graphics.addAll([firstGraphic, secondGraphic, thirdGraphic]);
```

What I'm getting at is, we don't want all that stuff jumbled up where the variable is declared, but also we don't want to declare it `late GraphicsOverlay _graphicsOverlay`.

So my idea is, it's declared like this:
```dart
final _graphicsOverlay = GraphicsOverlay();
```
... and in onMapViewReady, it gets the rest of the setup work done:
```dart
void onMapViewReady() async {
  _graphicsOverlay.renderer = SimpleRenderer(symbol: defaultMarkerSymbol)
  ..graphics.addAll([firstGraphic, secondGraphic, thirdGraphic]);
  // ...
}
```

Maybe I should phrase it: "If possible, avoid declaring a member variable `late`. Prefer instead to use the constructor when the variable is declared. But note that any other setup work should be done in `onMapViewReady()`."

shokhinaToj: Ohhh now it makes sense!

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
shokhinaToj: I believe so!

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
shokhinaToj: Wait @jenmerritt will we be in trouble for not following the commom design spec where the featureCount is just part of the bottom bar as plain text. When I was referring to iOS Sample and they have an alert dialog so I though that is more appropriate for a mob app.

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
shokhinaToj: Nope! it works without it just fine

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
jenmerritt: No the common design is desktop, we've implemented a better UI for mobile so this is fine in my opinion. The sample is emphasising the difference between a display filter and definition expression so I think showing the results in the dialog is a good pattern to follow.

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
```diff
@@ -27,48 +27,43 @@ class FilterByDefinitionExpressionOrDisplayFilterSample extends StatefulWidget {
...
-    map.operationalLayers.add(_featureLayer);
-    map.initialViewpoint = Viewpoint.withLatLongScale(
-      latitude: 37.7759,
-      longitude: -122.45044,
-      scale: 100000,
-    );
-
-    _mapViewController.arcGISMap = map;
-  }
+  var displayFilterDefinition;
```
jenmerritt, line 35:
```suggestion
  ManualDisplayFilterDefinition? displayFilterDefinition;
```
This is currently a dynamic, needs a type specifying

shokhinaToj: Okay! But shouldn't we be using type inference instead of specifying the type directly?

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
shokhinaToj: Okay, thanks!

lib/samples/filter_by_definition_expression_or_display_filter/filter_by_definition_expression_or_display_filter_sample.dart
jenmerritt: Yes - but it isn't smart enough to do that here, so it infers `dynamic` which we don't want, we want it to be a `ManualDisplayFilterDefinition`. 
If we were instantiating it here like `var displayFilterDefinition = ManualDisplayFilterDefinition()` then it would be ok.

PaulAllanSturm (approved):

PR #37, opened 2024-06-26
Apply common sample design to Apply simple renderer

kmueller-gis:
Updated the sample design based on the common sample document for Apply simple renderer to feature layer.
* Updated the sample code
* Udded metadata
* Updated the README to match the common text
* Added a screenshot


lib/samples/apply_simple_renderer_to_feature_layer/README.md
```diff
@@ -0,0 +1,34 @@
+# Apply simple renderer to feature layer
+
+Change the appearance of a feature layer with a renderer.
+
+![Image of map with features](apply_simple_renderer_to_feature_layer.png)
```
jenmerritt, line 5:
```suggestion
![Image of apply simple renderer to feature layer](apply_simple_renderer_to_feature_layer.png)
```

lib/samples/apply_simple_renderer_to_feature_layer/README.metadata.json
```diff
@@ -2,10 +2,22 @@
...
-    "redirect_from": [],
-    "relevant_apis": [],
-    "snippets": [],
+    "images": [
+        "apply_simple_renderer_to_feature_layer.png"
+    ],
+    "keywords": [
+        "feature layer",
+        "renderer",
+        "visualization"
```
jenmerritt, line 11: Keywords should contain both the relevant apis as well as the tags. There's a script you can run to generate this - I'll add the details in the issue.

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
```diff
@@ -42,11 +43,11 @@ class _ApplySimpleRendererToFeatureLayerSampleState
...
-        x: -9177250,
-        y: 4247200,
+        x: -9177343,
+        y: 4247283,
         spatialReference: SpatialReference.webMercator,
       ),
-      scale: 9500,
+      scale: 4750,
     );
     _mapViewController.arcGISMap = _map;
```
jenmerritt, line 52: We've established a pattern going forward that we'll do this setup within `_onMapViewReady`. It makes the story cleaner for those using the samples, which is sometimes a "quirk" of our samples - to balance absolute best practice with being very readable. Even though having it in `initState` is completely valid here. (see this discussion for background https://github.com/Esri/arcgis-maps-sdk-flutter-samples/pull/35#discussion_r1654467239)

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
```diff
@@ -85,11 +83,13 @@ class _ApplySimpleRendererToFeatureLayerSampleState
 
   void resetRenderer() {
     _featureLayer.resetRenderer();
```
jenmerritt, line 85:
```suggestion
    // reset the renderer on the feature layer
    _featureLayer.resetRenderer();
```
We're also adding in comments throughout the samples, again somewhat verbose in terms of best practice, but supports the workflows of reading through samples to all levels of developers both from a Flutter and ArcGIS perspective. Note the most recent PRs as examples (comments are missing from the original implementations)

jenmerritt (commented):
Thanks @kmueller-gis - the sample itself looks good, just some sample presentation specifics to follow (we'll improve our internal doc as we go too).

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
```diff
@@ -62,20 +63,17 @@ class _ApplySimpleRendererToFeatureLayerSampleState
                 controllerProvider: () => _mapViewController,
```
PaulAllanSturm, line 63: Another little tidbit -- we want to add `top: false` to SafeAreas when the SafeArea has a map view at the top. (When the sample is launched individually, there is no AppBar, so we want the map to run all the way up under the notch and what-not.)

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
PaulAllanSturm: We'll also need to add in the `setState(() => _ready = true)` mechanism, to protect the button callbacks from hitting the `late FeatureLayer _featureLayer` before it gets set.

Actually I really like using the `Visibility(visible: _ready, child: overlayControls(context))` pattern because it protects all the callbacks in one fell swoop. Maybe that can be a pattern for all samples that have `late` variables used in callbacks?

kmueller-gis: Added a Visibility widget with overlay controls that block user input.

jenmerritt: I quite like the use of the progress indicator, as this can be useful for samples that are downloading data too, and makes it clear something is happening. It is something noted in an issue about the offline data. I also used a progress indicator in show service area but looking at it again I think that sample could be improved with what you're saying @PaulAllanSturm, as I do `_ready` checks in multiple places, which is probably not necessary - I could probably capture that in a higher level widget.

So maybe we can make the pattern include a `CircularProgressIndicator`.

jenmerritt: Oh - and for `Visibility(visible: _ready, child: overlayControls(context))` - @PaulAllanSturm are you meaning actually using `overlayControls` as the general rule? I personally prefer using the widget tree directly unless it's getting really long. But definitely agree on the Visibility pattern.

PaulAllanSturm: For `CircularProgressIndicator`, that makes sense for when there's a download going on, plus for any long operation. As for the gap from creating the widget until the map is ready, it might be overkill, especially if it is visually unpleasant. There might be enough visual feedback already, as you go from a clear map to seeing the buttons appear.

`visible: _ready` is the important part of the `Visibility` pattern. `overlayControls()` would just be when we want to factor out a busy `build()` method.

jenmerritt: Sounds good to me 👍

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
```diff
@@ -27,69 +27,94 @@ class ApplySimpleRendererToFeatureLayerSample extends StatefulWidget {
...
-                  onPressed: overrideRenderer,
-                  child: const Text('Override'),
+                Expanded(
+                  child: ArcGISMapView(
+                    controllerProvider: () => _mapViewController,
+                    onMapViewReady: onMapViewReady,
+                  ),
                 ),
-                const Spacer(),
+                SizedBox(
```
jenmerritt, line 50: @kmueller-gis - maybe a good candidate for looking into the sized box discussion?

kmueller-gis: When I removed the height of 50, the box resized to fit the the button. This seems to work.

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
kmueller-gis: This page load happens so fast, we don't usually see the overlay controls. This does contain the `CircularProgressIndicator`. In the case that the overlay controls are around for a bit, it is good to have something moving indicating that the app didn't just lock up.

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
PaulAllanSturm: I don't actually know, but I wonder if SizedBox is the best Widget for the job here. Can it be done away with completely? Can the Center widget handle the necessary duties? Maybe there's some other flags we can pass to Center. Or maybe there's another widget that is simpler than SizedBox, something in the Container family.

Overall the goal is to figure out the simplest way to have a row of controls at the bottom, with a minimum height, that allows the Expanded around the MapView to take up all the slack.

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
PaulAllanSturm: Oh, in my head I was thinking of "overlay controls" to mean the buttons that operate the various functionality (they're overlaying the MapView). You were thinking of "overlay controls" as the CircularProgressIndicator that overlays the buttons.

Yeah if the onMapViewReady has any `awaits` for network resources, there should be a CircularProgressIndicator in there somehow. Maybe there's a more appropriate Widget than Visibility for that scenario. I think there is, but I'm not going to remember it until after I submit this comment.

PaulAllanSturm: Maybe IndexedStack https://api.flutter.dev/flutter/widgets/IndexedStack-class.html

PaulAllanSturm: I'd be happy to move this discussion to a separate issue so that it doesn't have to hold up the PR, if you agree.

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
```diff
@@ -27,69 +27,93 @@ class ApplySimpleRendererToFeatureLayerSample extends StatefulWidget {
...
-      ArcGISPoint(
-        x: -9177250,
-        y: 4247200,
-        spatialReference: SpatialReference.webMercator,
-      ),
-      scale: 9500,
-    );
-    _mapViewController.arcGISMap = _map;
-  }
+  final _mapViewController = ArcGISMapView.createController();
```
PaulAllanSturm, line 31: it's customary these days to put a comment before this line, like `  // create a map view controller`

kmueller-gis: All properties have been commented.

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
kmueller-gis: The `SizedBox` widget has been removed leaving only the `Center`. Still looks good.

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
kmueller-gis: I've created discussion #1908 in the Flutter repo to continue this conversation.

PaulAllanSturm (approved):

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
```diff
@@ -27,69 +27,95 @@ class ApplySimpleRendererToFeatureLayerSample extends StatefulWidget {
 
 class _ApplySimpleRendererToFeatureLayerSampleState
     extends State<ApplySimpleRendererToFeatureLayerSample> {
-  final _mapViewController = ArcGISMapView.createController();
+  // The feature layer that will host the symbolized features
   late FeatureLayer _featureLayer;
```
shokhinaToj, line 31: @PaulAllanSturm is this one of those places where we can avoid using `late` by initializing it with a constructor

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
```diff
@@ -27,69 +27,95 @@ class ApplySimpleRendererToFeatureLayerSample extends StatefulWidget {
...
               ],
-            )
+            ),
+            // Overlay controls to prevent interaction before state is ready
+            Visibility(
+              visible: !_ready,
+              child: SizedBox.expand(
+                child: Container(
+                  color: const Color.fromARGB(77, 255, 255, 255),
+                  child: const Center(child: CircularProgressIndicator()),
```
shokhinaToj, line 72: missing trailing comma?

lib/samples/apply_simple_renderer_to_feature_layer/apply_simple_renderer_to_feature_layer_sample.dart
kmueller-gis: Take a look at the latest change. I moved the instantiation from `onMapViewReady` to the property declaration. This design waits to instantiate the property until the first time it is used.

I still kept the `_ready` state just to make sure the map was all ready before interaction occured.

PaulAllanSturm: Member variables already have lazy initialization as a built-in part of Dart, so there's no need to do the `() {...}()` trick. It could just be:

Option A
```dart
final _featureLayer = FeatureLayer.withFeatureTable(
  ServiceFeatureTable.withUri(
    Uri.parse(
      'https://services.arcgis.com/V6ZHFr6zdgNZuVG0/arcgis/rest/services/Landscape_Trees/FeatureServer/0'),
  ),
);
```

The other option is

Option B
```dart
  late final FeatureLayer _featureLayer;

  // ...

  void onMapViewReady() {
    final serviceFeatureTable = ServiceFeatureTable.withUri(
      Uri.parse(
        'https://services.arcgis.com/V6ZHFr6zdgNZuVG0/arcgis/rest/services/Landscape_Trees/FeatureServer/0'),
    );
    _featureLayer = FeatureLayer.withFeatureTable(serviceFeatureTable);
  }
```

The competing principles are:

Principle A: `late` is hazardous
Principle B: Samples should read as heavily-commented, top-down step-by-step tutorials

Unfortunately the principles are in conflict, because there's no way to avoid `late` without having a bunch of important stuff crammed up in the initializer, out of line with the flow of the tutorial. So I think we have to go with Principle B / Option B in this case.

The general rule would be: declare member variables by calling their constructor if the call is simple; otherwise, declare it `late` and initialize it in `onMapViewReady()`. Once we get into a constructor nested inside a constructor, it's no longer simple.

kmueller-gis: I concur. I'll move the init back to `onMapViewReady`.

jenmerritt (approved):
Hate to be a pain but could we get an updated screengrab with the new button 😅  Otherwise I'm approving with just a couple of doc suggestions so as to not hold this up waiting for me since I'm at the end of my day.

PR #54, opened 2024-07-08
"Show Portal User Info" sample

PaulAllanSturm:
Implemented the "Show Portal User Info" sample. The "How it works" and "Relevant API" sections were changed from the design to reflect the new authentication mechanism (ArcGISAuthenticationChallengeHandler).

I updated `generate_samples_list.dart` so that it pretty-prints the JSON, rather than putting everything on a single line. This should prevent most merge conflicts on the `generated_samples_list.json` file. 

When comparing to the `authenticate_with_oauth_sample.dart` sample, I noticed that it was calling `super.dispose()` in the wrong place.


lib/samples/show_portal_user_info/show_portal_user_info_sample.dart
```diff
@@ -0,0 +1,168 @@
...
+  }
+}
+
+extension on LoadableImage {
+  // A helper method to load the image bytes.
+  Future<Uint8List> loadBytes() async {
+    await load();
+    return image!.getEncodedBuffer();
+  }
+}
```
jenmerritt, line 168: Is it qorth a discussion if we provide this in the API if it's going to be required in most use cases to save the extension being needed per application?

jenmerritt: Nice, it'll clean it up like we do for the basemap and tile cache thumbnails 👍

jenmerritt (approved):
Maybe just check with others re: screenshot but otherwise LGTM

kmueller-gis (approved):

PR #56, opened 2024-07-09
"Generate Offline Map" sample

PaulAllanSturm:
Implemented "Generate Offline Map" sample.


assets/generated_samples_list.json
jenmerritt, line 1: Just to say this is hugely improved after your previous change to the script, nice! 🎉

jenmerritt (commented):
My thoughts on this sample would be:
- We don't need a reset button
- We should add a cancel button
- I really don't mind how we show progress, the way you have it is fine as far as I'm concerned, but the other platforms have a progress bar with % by it so maybe we should follow suit.

See these more recent implementations - [Swift](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/main/Shared/Samples/Generate%20offline%20map) / [Kotlin](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/blob/main/generate-offline-map/src/main/java/com/esri/arcgismaps/sample/generateofflinemap/MainActivity.kt) / [MAUI](https://github.com/Esri/arcgis-maps-sdk-dotnet-samples/tree/main/src/MAUI/Maui.Samples/Samples/Map/GenerateOfflineMap)

lib/samples/generate_offline_map/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+![Image of generate offline map](generate_offline_map.png)
+
+## Use case
+
+Taking a web map offline allows users continued productivity when their network connectivity is poor or nonexistent. For example, by taking a map offline, a field worker inspecting utility lines in remote areas could still access a feature's location and attribute information.
+
+## How to use the sample
+
+Once the map loads, zoom to the extent you want to take offline. The red border shows the extent that will be downloaded. Click the "Take Map Offline" button to start the offline map job. The progress indicator will show the job's progress. When complete, the offline map will replace the online map in the map view.
```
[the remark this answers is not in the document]
PaulAllanSturm: fixed

PaulAllanSturm:
> My thoughts on this sample would be:
> 
> * We don't need a reset button
> * We should add a cancel button
> * I really don't mind how we show progress, the way you have it is fine as far as I'm concerned, but the other platforms have a progress bar with % by it so maybe we should follow suit.
> 
> See these more recent implementations - [Swift](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/main/Shared/Samples/Generate%20offline%20map) / [Kotlin](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/blob/main/generate-offline-map/src/main/java/com/esri/arcgismaps/sample/generateofflinemap/MainActivity.kt) / [MAUI](https://github.com/Esri/arcgis-maps-sdk-dotnet-samples/tree/main/src/MAUI/Maui.Samples/Samples/Map/GenerateOfflineMap)

fixed

lib/samples/generate_offline_map/generate_offline_map_sample.dart
```diff
@@ -0,0 +1,227 @@
...
+  State<GenerateOfflineMapSample> createState() =>
+      _GenerateOfflineMapSampleState();
+}
+
+class _GenerateOfflineMapSampleState extends State<GenerateOfflineMapSample>
+    with SampleStateSupport {
+  // Create a controller for the map view.
+  final _mapViewController = ArcGISMapView.createController();
+  // Create a Graphic to show the area to be taken offline.
+  final _regionGraphic = Graphic();
```
jenmerritt, line 38: I think we need to use a UI component for this. When you rotate the map view right now, the graphic red box rotates around and changes size whereas I think it should be fixed. Other platforms have used `Rectangle` API equivalents to fix it in place. There is a `Rect` class in Dart but maybe `SizedBox` or `Positioned` widgets would work. Before we change it though I've posed the question to the other SDKs to confirm, and will also update the design if this is the case to make it less vague.

lib/samples/generate_offline_map/generate_offline_map_sample.dart
```diff
@@ -0,0 +1,227 @@
...
+              ],
+            ),
+            // Display a progress indicator and prevent interaction until state is ready.
+            Visibility(
+              visible: !_ready,
+              child: SizedBox.expand(
+                child: Container(
+                  color: Colors.white30,
+                  child: const Center(
+                    child: CircularProgressIndicator(),
```
jenmerritt, line 87: I think from the skeleton we can remove the comma here

jenmerritt (commented):
Changes look good. I just want to clarify the design around the red box behavior as per my comment below.

lib/samples/generate_offline_map/generate_offline_map_sample.dart
PaulAllanSturm: The Kotlin version uses a Graphic and viewpointChanged the same way: https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/blob/74a3c10646b5f5e39644c995fc1e28b3cb9515f0/generate-offline-map/src/main/java/com/esri/arcgismaps/sample/generateofflinemap/MainActivity.kt#L71C17-L71C29

I tried using SizedBox/Positioned but it gets really hairy trying to map a Widget's on-screen position to the map view. It might be possible, but you end up with a bunch of code like our internal `ViewMetrics.fromContext`. It would be a big detour from the tutorial flow.

jenmerritt: Yeah the Kotlin sample has the same behavior issues as well though. To me it's not that easy to use right now because as soon as you start interacting with the map view you lose your rectangle. Maybe we can disable map view rotation, might improve the behavior if it's not a simple thing to achieve with a widget in Flutter? Agree we want to get the balance in the sample.

PaulAllanSturm: I re-worked it with the fixed Widget outline and the screen-to-map conversion logic.

shokhinaToj (approved):
Looks great!

PR #63, opened 2024-07-11
Update redirect for OAuth

jenmerritt:
Updates to a unique OAuth redirect to avoid clash with other SDK sample viewers on same device. I created a separate OAuth credential so also updates the clientId.


jenmerritt:
I could move the dispose calls in this PR if it's easier to pull those out of generate offline map @PaulAllanSturm

PaulAllanSturm (approved):

PaulAllanSturm:
> I could move the dispose calls in this PR if it's easier to pull those out of generate offline map @PaulAllanSturm

I don't think we'll run into merge conflicts so it should all be fine as it is now.

PR #64, opened 2024-07-11
Show Legend Sample

shokhinaToj:
Current Flutter sample: https://devtopia.esri.com/runtime/common-samples/tree/main/designs/ShowLegend

Implemented a new sample according to common design.


lib/samples/show_legend/README.md
```diff
@@ -0,0 +1,33 @@
...
+3. The names and images are then displayed next to each other in a list view UI control.
+
+## Relevant API
+
+* LayerContent
+* LegendInfo
+
+## Additional information
+
+This sample uses the legend toolkit component, which requires the toolkit to be cloned and set up locally. Visit the repository for information about setting up the toolkit, platform availabilty, as well as code for the underlying component.
```
PaulAllanSturm, line 29: We should remove the "Additional information" section until we have the toolkit component available.

lib/samples/show_legend/show_legend_sample.dart
```diff
@@ -0,0 +1,174 @@
...
+
+class ShowLegendSample extends StatefulWidget {
+  const ShowLegendSample({super.key});
+  @override
+  State<ShowLegendSample> createState() => _ShowLegendSampleState();
+}
+
+class _ShowLegendSampleState extends State<ShowLegendSample>
+    with SampleStateSupport {
+  // create a map view controller
```
PaulAllanSturm, line 30: We decided to standardize comments with a capital letter and a period --
```dart
// Create a map view controller.
```
We should apply that to the comments throughout.

lib/samples/show_legend/show_legend_sample.dart
```diff
@@ -0,0 +1,174 @@
...
+              // add a map view to the widget tree and set a controller.
+              child: ArcGISMapView(
+                controllerProvider: () => _mapViewController,
+                onMapViewReady: onMapViewReady,
+              ),
+            ),
+            // add a dropdown button to the widget tree.
+            DropdownButtonHideUnderline(
+              child: DropdownButton(
+                value: _selectedLegend,
```
PaulAllanSturm, line 59: I think you can just delete this line and delete where `_selectedLegend` is declared further up. The `value` parameter doesn't have to be specified.

lib/samples/show_legend/show_legend_sample.dart
```diff
@@ -0,0 +1,174 @@
...
+                hint: const Text(
+                  'Legend',
+                  style: TextStyle(
+                    color: Colors.deepPurple,
+                  ),
+                ),
+                elevation: 16,
+                style: const TextStyle(color: Colors.deepPurple),
+                // no need to set up onChanged callback
+                onChanged: (n) {},
```
PaulAllanSturm, line 71: The usual pattern is to use `_` as the name of unused parameters, so this would be: `onChanged: (_) {},`

lib/samples/show_legend/show_legend_sample.dart
```diff
@@ -0,0 +1,174 @@
...
+          'https://sampleserver6.arcgisonline.com/arcgis/rest/services/Census/MapServer'),
+    );
+    // create a feature table
+    final featureTable = ServiceFeatureTable.withUri(
+      Uri.parse(
+          'https://sampleserver6.arcgisonline.com/arcgis/rest/services/Recreation/FeatureServer/0'),
+    );
+    // create a feature layer
+    final featureLayer = FeatureLayer.withFeatureTable(featureTable);
+    // add image and feature layers to the opertaional layers list of the map
```
PaulAllanSturm, line 96: typo "opertaional" should be "operational"

lib/samples/show_legend/show_legend_sample.dart
```diff
@@ -0,0 +1,174 @@
...
+            style: const TextStyle(
+              fontSize: 15,
+              fontWeight: FontWeight.bold,
+            ),
+          ),
+        ),
+      );
+
+      // add current layer's legends to the dropdown list
+      for (var legend in layerLegends) {
```
PaulAllanSturm, line 128: I think this can be `final legend` instead of `var legend`

lib/samples/show_legend/show_legend_sample.dart
```diff
@@ -0,0 +1,174 @@
...
+        try {
+          // create a swatch for the legend
+          _arcGISImage = await legend.symbol!.createSwatch(
```
PaulAllanSturm, line 131: `_arcGISImage` shouldn't be a member variable. It should just be a temporary local variable declared right before the "try" (and not have an underscore).

lib/samples/show_legend/show_legend_sample.dart
```diff
@@ -0,0 +1,174 @@
...
+            screenScale: screenScale,
+            backgroundColor: Colors.transparent,
+            size: const Size.square(6),
+          );
+        } catch (e) {}
```
PaulAllanSturm, line 136: I'm not sure about this empty "catch" block. What exceptions might show up here? Maybe we don't need a try/catch at all.

lib/samples/show_legend/show_legend_sample.dart
PaulAllanSturm: And instead of using `!` with `symbol!`, it should check if `symbol` is `null`. If it's `null`, it should handle having a null ArcGISImage down below. Instead of doing `_arcGISImage!.getEncodedBuffer()`, it should check if `_arcGISImage` is null and maybe substitute an Icon or something.

lib/samples/show_legend/show_legend_sample.dart
Shokhinabonu: I can remove the value param from the DropdownButton widget but if I remove it from the DropdownMenuItem widgets it crashes. Is it fine if we pass value for each DropdownMenuItem but not for DropdownButton?

PaulAllanSturm: Yeah I think that's fine.

lib/samples/show_legend/show_legend_sample.dart
Shokhinabonu: Since it returns a future, I though we should handle a case if it returns an error. But if I am adding all these null checkers, would try/catch be unnecessary ?

PaulAllanSturm: Yeah after the null checks are in, the try/catch will be unnecessary.

lib/samples/show_legend/show_legend_sample.dart
```diff
@@ -0,0 +1,175 @@
...
+        final ArcGISImage? arcGISImage;
+        // Create a swatch for the legend if the legend exists.
+        if (legend.symbol != null) {
+          arcGISImage = await legend.symbol!.createSwatch(
+            screenScale: screenScale,
+            backgroundColor: Colors.transparent,
+            size: const Size.square(6),
+          );
+        } else {
+          arcGISImage = null;
```
PaulAllanSturm, line 133: I think you don't need this `else` block. `arcGISImage` defaults to `null` automatically.

lib/samples/show_legend/show_legend_sample.dart
```diff
@@ -0,0 +1,175 @@
...
+    _map.initialViewpoint = Viewpoint.fromCenter(
+      ArcGISPoint(
+        x: -11e6,
+        y: 6e6,
+        spatialReference: SpatialReference.webMercator,
+      ),
+      scale: 9e7,
+    );
+    // Reset the state once the dropdown list is updated.
+    setState(() {});
```
PaulAllanSturm, line 173: I would put this right after the big block of code that builds `_legendsDropDown`, to make it a little more clear what it's doing.

Actually, another idea would be to declare a local temporary variable, `final legendsDropDown = <DropdownMenuItem<LegendInfo>>[]`, and use that one to build up the list. Then, after the whole list is built, do a call to `setState` like:
```dart
setState(() => _legendsDropDown = legendsDropDown);
```
That way, we don't have to worry about the list getting built a little at a time with "async" jumps in between. I'm worried that the `build()` method might get called while the dropdown list is only partially built.

lib/samples/show_legend/show_legend_sample.dart
shokhinaToj: If I remove it, it says `The final variable 'arcGISImage' can't be read because it's potentially unassigned at this point.
Ensure that it is assigned on necessary execution paths.`

PaulAllanSturm: Oh, maybe that's because it's declared as `final ArcGISImage? arcGISImage`. What happens if you declare it as just `ArcGISImage? arcGISImage`

lib/samples/show_legend/show_legend_sample.dart
shokhinaToj: Yes that sounds like a good idea - we would also avoid having an empty setState()

lib/samples/show_legend/show_legend_sample.dart
shokhinaToj: That works!

PaulAllanSturm (approved):

lib/samples/show_legend/README.md
```diff
@@ -0,0 +1,29 @@
+# Show legend
+
+Build a legend for all the operational layers in the map.
+
+![Image of build legend](show_legend.png)
```
jenmerritt, line 5:
```suggestion
![Image of show legend](show_legend.png)
```

lib/samples/show_legend/README.md
```diff
@@ -0,0 +1,29 @@
...
+3. The names and images are then displayed next to each other in a list view UI control.
+
+## Relevant API
+
+* LayerContent
+* LegendInfo
+
+## Tags
+
+legend, legend info, symbol swatch, toolkit
```
jenmerritt, line 29:
```suggestion
legend, legend info, symbol swatch
```
We should remove the toolkit tag until we have it

lib/samples/show_legend/README.metadata.json
```diff
@@ -0,0 +1,25 @@
...
+    "description": "Build a legend for all the operational layers in the map.",
+    "ignore": false,
+    "images": [
+        "show_legend.png"
+    ],
+    "keywords": [
+        "legend",
+        "legend info",
+        "symbol swatch",
+        "toolkit",
```
jenmerritt, line 12:
```suggestion
```
This will be removed when regenerated after updating the readme.

lib/samples/show_legend/show_legend_sample.dart
```diff
@@ -0,0 +1,174 @@
...
+    final featureTable = ServiceFeatureTable.withUri(
+      Uri.parse(
+          'https://sampleserver6.arcgisonline.com/arcgis/rest/services/Recreation/FeatureServer/0'),
+    );
+    // Create a feature layer.
+    final featureLayer = FeatureLayer.withFeatureTable(featureTable);
+    // Add image and feature layers to the operational layers list of the map.
+    _map
+      ..operationalLayers.add(imageLayer)
+      ..operationalLayers.add(featureLayer);
```
jenmerritt, line 94:
```suggestion
    _map.operationalLayers.addAll([imageLayer, featureLayer]);
```

PR #66, opened 2024-07-12
cs/Update Show-grid Sample

changanxian:
Based on the test and feedback of the Sprint Review meeting, Here are some changes made:
- Use enum to hold the list of constant property values
- The changed grid property status is preserved in the UI
- The barrier color is changed to transparent for users to see the changes easily
- The dialog is moved to the bottom center, making more room to view for ArcGISMapView

### DIFF
--- a/lib/samples/show_grid/show_grid_sample.dart
+++ b/lib/samples/show_grid/show_grid_sample.dart
@@ -36,6 +36,13 @@ class _ShowGridSampleState extends State<ShowGridSample>
     spatialReference: SpatialReference.webMercator,
   );
 
+  var _gridType = GridType.latitudeLongitude;
+  var _gridColorType = GridColorType.red;
+  var _gridLabelColorType = GridColorType.red;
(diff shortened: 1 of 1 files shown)


lib/samples/show_grid/show_grid_sample.dart
```diff
@@ -351,80 +386,113 @@ class _GridOptionsState extends State<GridOptions> with SampleStateSupport {
...
-      case 'Green':
-        return Colors.green;
-      case 'Yellow':
-        return Colors.yellow;
-      default:
-        return Colors.red;
-    }
+  // Set the visibility of the label format dropdown based on the grid type.
+  bool _shouldShowLabelFormat(GridType grid) {
+    return (grid == GridType.latitudeLongitude) ? true : false;
```
PaulAllanSturm, line 401: This could more simply be `return grid == GridType.latitudeLongitude;`

PaulAllanSturm (approved):

PR #67, opened 2024-07-15
Search functionality 

shokhinaToj:
Adding search bar to the main sample viewer screen

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -43,21 +43,23 @@ class SampleViewerApp extends StatefulWidget {
 
 class _SampleViewerAppState extends State<SampleViewerApp> {
   final _allSamples = <Sample>[];
+  final _searchFocusNode = FocusNode();
+  final _textEditingController = TextEditingController();
   var _filteredSamples = <Sample>[];
   bool _ready = false;
   bool _searchHasFocus = false;
-  final _searchFocusNode = FocusNode();
-  final _textEditingController = TextEditingController();
 
   @override
   void initState() {
     super.initState();
     loadSamples();
-    _searchFocusNode.addListener(() {
-      if (_searchFocusNode.hasFocus != _searchHasFocus) {
-        setState(() => _searchHasFocus = _searchFocusNode.hasFocus);
-      }
-    });
+    _searchFocusNode.addListener(
+      () {
+        if (_searchFocusNode.hasFocus != _searchHasFocus) {
+          setState(() => _searchHasFocus = _searchFocusNode.hasFocus);
+        }
+      },
+    );
   }
 
   @override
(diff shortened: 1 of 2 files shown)


lib/main.dart
```diff
@@ -67,20 +83,89 @@ class _SampleViewerAppState extends State<SampleViewerApp> {
...
     final sampleData = jsonDecode(jsonString);
     for (final s in sampleData.entries) {
-      _samples.add(Sample.fromJson(s.value));
+      _allSamples.add(Sample.fromJson(s.value));
     }
+    _filteredSamples = _allSamples;
     setState(() => _ready = true);
   }
+
+  void onSearchChanged(String searchText) {
```
JamesMBallard, line 144: Can this method be optimized? Is there any way to avoid jumping between a set and a list?

lib/main.dart
```diff
@@ -67,20 +83,89 @@ class _SampleViewerAppState extends State<SampleViewerApp> {
         appBar: AppBar(
           title: const Text(title),
         ),
-        body: _ready
-            ? SampleListView(samples: _samples)
-            : const Center(child: Text('Loading samples...')),
+        body: Listener(
+          onPointerDown: (_) => FocusManager.instance.primaryFocus?.unfocus(),
```
PaulAllanSturm, line 87: I'm not sure, but I think `onPointerDown` shouldn't be necessary. What is it trying to do? Is this to unfocus the search field and dismiss the keyboard? If it really is necessary, at least it should move down the widget tree and be closer to the `SampleListView`. Otherwise when you tap in the TextField, it immediately tries to unfocus it and then focuses it.

lib/models/sample.dart
```diff
@@ -26,13 +26,17 @@ class Sample {
   final List<String> _snippets;
   final String _title;
   final Widget _sampleWidget;
+  final List<String> _keywords;
+  final List<String> _relevantApis;
```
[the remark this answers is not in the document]
jenmerritt: You should actually only need keywords too. The keywords metadata property contains both relevant API and tags from the readme.

lib/main.dart
```diff
@@ -67,20 +83,89 @@ class _SampleViewerAppState extends State<SampleViewerApp> {
...
+                    ),
+            ],
+          ),
+        ),
       ),
     );
   }
 
+  void onSearchSuffixPressed() {
+    if (!_searchHasFocus) {
```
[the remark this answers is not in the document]
shokhinaToj: That's what I thought but it only worked when it was reversed - however, I think that Listener at the top was altering the focus on the page so when I took it out it works as expected. So this is fixed!

lib/main.dart
PaulAllanSturm: Could it wrap the Expanded that wraps the SampleListView? Since it is part of the search mechanism, I'd prefer to keep it all together here in this file. But otherwise putting it inside SampleListView would be acceptable.

lib/main.dart
```diff
@@ -67,20 +83,89 @@ class _SampleViewerAppState extends State<SampleViewerApp> {
...
     for (final s in sampleData.entries) {
-      _samples.add(Sample.fromJson(s.value));
+      _allSamples.add(Sample.fromJson(s.value));
     }
+    _filteredSamples = _allSamples;
     setState(() => _ready = true);
   }
+
+  void onSearchChanged(String searchText) {
+    List<Sample> results = [];
```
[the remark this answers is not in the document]
shokhinaToj: I agree!

lib/main.dart
shokhinaToj: As Paul suggested I think there is no need for Set anymore since now I am just iterating through the samples only once. So I will just get rid of it!

lib/main.dart
shokhinaToj: Actually I just tried wrapping the SampleListView directly instead of the Expanded and this isn't throwing any exceptions!

PaulAllanSturm (approved):

lib/main.dart
```diff
@@ -42,13 +42,31 @@ class SampleViewerApp extends StatefulWidget {
 }
 
 class _SampleViewerAppState extends State<SampleViewerApp> {
-  final _samples = <Sample>[];
+  final _allSamples = <Sample>[];
+  final _searchFocusNode = FocusNode();
+  final _textEditingController = TextEditingController();
+  var _filteredSamples = <Sample>[];
   bool _ready = false;
+  bool _searchHasFocus = false;
```
JamesMBallard, line 50: Is this needed? Why do we need both `_searchHasFocus` and `_searchFocusNode` which has a `hasFocus` property?

PaulAllanSturm: It's not strictly necessary, but the problem is that `_searchFocusNode` can change its `hasFocus` depending on user actions, and that doesn't in and of itself cause a rebuild of the widget tree. So we need to listen for changes to `_searchFocusNode` to detect when to call `setState()` (which causes the actual rebuild). That requires just a small bit of logic down below in `_searchFocusNode.addListener(...)`.

JamesMBallard: In that case can we have `_searchFocusNode.addListener` call setState without needing to duplicate that property? Or, only do that somehow when hasFocus changed without needing to maintain a separate bool?

If not, no problem.

PaulAllanSturm: The `addListener` call is over-inclusive. There are other changes that can trigger the `addListener` callback, but we only want to call `setState()` if `hasFocus` changes. As far as I know, there's no way to get a notification for only `hasFocus`. So we have to listen for everything, and from there check for a change specifically to `hasFocus`.

The other nice thing is that `setState()` is self-documenting, in that you can see what property is being changed. If you did just a generic empty `setState(() {})`, you kinda lose the thread of what's going on.

JamesMBallard (approved):

PR #68, opened 2024-07-16
Escape single quotes in user input

PaulAllanSturm:
The 'single quote' is a special character in SQL, so we have to replace them with a pair of single quotes when using user input as part of a `whereClause`.

Also, I reviewed the other SDKs versions of this sample. They use `%` to allow substring matches. So I added a `%` at the end of our query, so that e.g., `Calif` matches `California` and so forth.

### DIFF
--- a/lib/samples/query_feature_table/query_feature_table_sample.dart
+++ b/lib/samples/query_feature_table/query_feature_table_sample.dart
@@ -123,8 +123,8 @@ class _QueryFeatureTableSampleState extends State<QueryFeatureTableSample>
 
     // Create query parameters and set the where clause.
     final queryParameters = QueryParameters();
-    final stateName = value.trim().toUpperCase();
-    queryParameters.whereClause = "upper(STATE_NAME) LIKE '$stateName'";
+    final stateName = value.trim().toUpperCase().replaceAll("'", "''");
+    queryParameters.whereClause = "UPPER(STATE_NAME) LIKE '$stateName%'";
 
     // Query the feature table with the query parameters.
     final queryResult =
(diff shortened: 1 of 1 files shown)


lib/samples/query_feature_table/query_feature_table_sample.dart
```diff
@@ -123,8 +123,8 @@ class _QueryFeatureTableSampleState extends State<QueryFeatureTableSample>
 
     // Create query parameters and set the where clause.
     final queryParameters = QueryParameters();
-    final stateName = value.trim().toUpperCase();
-    queryParameters.whereClause = "upper(STATE_NAME) LIKE '$stateName'";
+    final stateName = value.trim().toUpperCase().replaceAll("'", "''");
```
changanxian, line 126: why do you want to replace it with a pair of single quotes instead of removing it `replaceAll("'", "")`

PaulAllanSturm: Removing single quotes would certainly fix this issue, because we know that no state names have single quotes. But as a more general solution, we should escape single quotes in case the user intentionally wants to search for something that has single quotes. For example, if they were searching for names, they may want to specify "O'Brian". I realize that's not a concern here, but I think our sample code should do it this way in case people copy our code.

changanxian: Oh, good points.

changanxian (approved):

lib/samples/query_feature_table/query_feature_table_sample.dart
jenmerritt: I think we should add a comment here to explain that. Something like: 
`// Replace single quotes with a pair of single quotes to form a valid SQL query.`

Since it's specific to the SQL query - otherwise it might raise the question of why not just use a validator, whereas they are valid, it's just we have to meet the SQL reqs. If that makes any sense.

changanxian (approved):

jenmerritt (approved):
👍

PR #76, opened 2024-07-19
Set Basemap Sample

shokhinaToj:
Current Flutter sample: https://github.com/ArcGIS/arcgis-maps-sdk-flutter-samples/blob/main/lib/src/samples/set_basemap_sample.dart

Applie common sample design to this sample

Added comments, new README files, and a screenshot


shokhinaToj:
@jenmerritt for some reason the last two basemap options look odd - one of them is missing a thumbnail and other doesn't have a title. Is it supposed to be like this? (the screenshot for comparison is missing from common sample doc :(

PaulAllanSturm:
The hard-coded list of basemaps is probably not quite correct. That's why a few of them are odd.

We probably want to fetch the developer basemaps from `Portal.arcGISOnline()`. And I think that means we have to add a method to `Portal` that is currently unimplemented -- we have the `basemaps()` method, but the api_specification also defines a `developerBasemaps()` method that is what we want.

lib/samples/set_basemap/README.md
```diff
@@ -0,0 +1,29 @@
+# Set basemap
+
+Change a map's basemap. A basemap is beneath all layers on an `ArcGISMap` and is used to provide visual reference for the operational layers.
+
+![Image of change basemap](set_basemap.png)
```
PaulAllanSturm, line 5: This should be "Image of set basemap"

lib/samples/set_basemap/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Use case
+
+Basemaps should be selected contextually. For example, in maritime applications, it would be more appropriate to use a basemap of the world's oceans as opposed to a basemap of the world's streets.
+
+## How to use the sample
+
+Use the drop down menu to select the active basemap from the list of available basemaps.
```
PaulAllanSturm, line 13: "drop down menu" isn't right for mobile -- it should say "drawer" I think

lib/samples/set_basemap/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## How it works
+
+1. Create an `ArcGISMap` object.
+2. Set the map to the `ArcGISMapView` object.
```
PaulAllanSturm, line 18: It should be "the `ArcGISMapViewController` object"

lib/samples/set_basemap/README.md
```diff
@@ -0,0 +1,29 @@
...
+3. Choose a new basemap type and set it on the map.
+
+## Relevant API
+
+* ArcGISMap
+* ArcGISMapView
```
PaulAllanSturm, line 24: Should be `ArcGISMapViewController` -- which means gotta run the script after to generate the other metadata file.

PaulAllanSturm (commented):
We'll have to decide what set of basemaps we want to use. We added some odd ones to test various other things that probably no longer relevant to this sample.

jenmerritt:
> for some reason the last two basemap options look odd - one of them is missing a thumbnail and other doesn't have a title.

This was originally testing our Basemap API and some of the image APIs, so doesn't follow the common design. There were deliberately some basemaps that use their thumbnails and at least one basemap that doesn't have an image (and uses a default image asset). So ignore what is there already. As @PaulAllanSturm says, it's the developer basemaps we want. If you look at other platform implementations, they use the `BasemapGallery` toolkit component, which default to that list. We don't have the basemap toolkit component, so will just need to manually display that list for now.

PaulAllanSturm:
I'm adding Portal.developerBasemaps to the library so we can use it for this sample.

shokhinaToj:
I removed all the URIs to the basemaps from before and now using portal.developerBasemaps() but the last basemap thumbnail still has a missing title
Screenshot 2024-07-22 at 2 38 37 PM

PaulAllanSturm:
> I removed all the URIs to the basemaps from before and now using portal.developerBasemaps() but the last basemap thumbnail still has a missing title @jenmerritt @PaulAllanSturm 

Yeah just delete everything that mentions SanFrancisco or tiles. That's the one without the thumbnail. It's not part of the official design.

lib/samples/set_basemap/set_basemap_sample.dart
```diff
@@ -47,7 +51,7 @@ class _SetBasemapSampleState extends State<SetBasemapSample>
   @override
   void initState() {
     super.initState();
-
+    // Set the initial viewpoint and basemap for the map.
     _arcGISMap.initialViewpoint = _sanFranciscoViewpoint;
     _mapViewController.arcGISMap = _arcGISMap;
     _loadBasemapsFuture = loadBasemaps();
```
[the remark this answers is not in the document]
shokhinaToj: I tried this but for some reason when you put it in onMapViewReady it never loads the basemaps. Maybe because it builds the Drawer first before it reaches the onMapViewReady call?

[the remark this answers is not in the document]
PaulAllanSturm: You might be right. Probably `_loadBasemapsFuture` has to remain in `initState()`. So yeah, keep `initState()` around just for the `_loadBasemapsFuture` line. The other stuff can move to `onMapViewReady()`.

lib/samples/set_basemap/set_basemap_sample.dart
```diff
@@ -32,42 +30,45 @@ class SetBasemapSample extends StatefulWidget {
...
 
   @override
   Widget build(BuildContext context) {
     return Scaffold(
+      // Create a scaffold with a key to access the scaffold state.
       key: _scaffoldStateKey,
+      // Create an end drawer to display basemaps.
       endDrawer: Drawer(
         child: SafeArea(
+          top: false,
```
PaulAllanSturm, line 59: I think we need to remove `top: false`. We only use `top: false` so that the mapview runs all the way up to the top of the screen, but this SafeArea is for a different purpose. This SafeArea isn't covering the map widgets, just the Drawer. So for the Drawer, I think we just need a regular SafeArea with no changes.

lib/samples/set_basemap/set_basemap_sample.dart
```diff
@@ -115,33 +121,25 @@ class _SetBasemapSampleState extends State<SetBasemapSample>
     );
   }
 
+  void _onMapViewReady() {
```
PaulAllanSturm, line 124: This should be `onMapViewReady` (no _underscore).

lib/samples/set_basemap/set_basemap_sample.dart
```diff
@@ -115,33 +121,25 @@ class _SetBasemapSampleState extends State<SetBasemapSample>
...
+    // Set the map view controller's map to the ArcGIS map.
+    _mapViewController.arcGISMap = _arcGISMap; 
+  }
+
   void updateMap(Basemap basemap) {
+    // Update the map view with the selected basemap.
     _arcGISMap.basemap = basemap;
-    _mapViewController.setViewpointAnimated(_sanFranciscoViewpoint);
+    // Set the rotation angle to 0.
     _mapViewController.setViewpointRotation(angleDegrees: 0.0);
```
PaulAllanSturm, line 133: I'm not sure there's a point to setting the rotation back to 0.0. I think that was just part of the old San Francisco tile map stuff. I think we should get rid of the call to setViewpointRotation.

lib/samples/set_basemap/set_basemap_sample.dart
```diff
@@ -115,33 +121,25 @@ class _SetBasemapSampleState extends State<SetBasemapSample>
...
-      Basemap.withUri(Uri.parse(
-          'https://www.arcgis.com/home/item.html?id=2e8a3ccdfd6d42a995b79812b3b0ebc6')), // Outdoor
-      Basemap.withUri(Uri.parse(
-          'https://www.arcgis.com/home/item.html?id=7e2b9be8a9c94e45b7f87857d8d168d6')), // Streets night
-      Basemap.withStyle(BasemapStyle.arcGISNavigation), // No thumbnail
-    ]);
+    // Create a portal to access online items.
+    final portal = Portal.arcGISOnline();
+    // Load basemaps from portal.
+    List<Basemap> basemaps = await portal.developerBasemaps();
```
PaulAllanSturm, line 140: This could just be `final basemaps = `

lib/samples/set_basemap/set_basemap_sample.dart
shokhinaToj: Aaa okay that makes sense!

lib/samples/set_basemap/set_basemap_sample.dart
```diff
@@ -115,33 +120,23 @@ class _SetBasemapSampleState extends State<SetBasemapSample>
...
-      Basemap.withStyle(BasemapStyle.arcGISNavigation), // No thumbnail
-    ]);
+    // Create a portal to access online items.
+    final portal = Portal.arcGISOnline();
+    // Load basemaps from portal.
+    final basemaps = await portal.developerBasemaps();
 
-    // load each basemap to access and display attribute data in the UI
+    // Load each basemap to access and display attribute data in the UI.
     for (var basemap in basemaps) {
```
PaulAllanSturm, line 140: This could be `for (final basemap in basemaps) {`
And below `var thumbnail =` can be `final thumbnail =`

PaulAllanSturm (commented):
sorry one last thing

lib/samples/set_basemap/README.md
```diff
@@ -0,0 +1,29 @@
+# Set basemap
+
+Change a map's basemap. A basemap is beneath all layers on an `ArcGISMap` and is used to provide visual reference for the operational layers.
```
jenmerritt, line 3:
```suggestion
Change a map's basemap. A basemap is beneath all layers on a map and is used to provide visual reference for the operational layers.
```
The design is wrong here, we shouldn't have classes referenced in the description. So just tweak to map. (I'll clean up the design).

lib/samples/set_basemap/README.metadata.json
```diff
@@ -1,11 +1,25 @@
 {
     "category": "Maps",
-    "description": "Set a map's basemap. A basemap is beneath all layers on a map and is used to provide visual reference for the operational layers.",
+    "description": "Change a map's basemap. A basemap is beneath all layers on an `ArcGISMap` and is used to provide visual reference for the operational layers.",
```
jenmerritt, line 3: This will need regenerating with that change

jenmerritt (commented):
Just a minor doc thing from me! I've not done a second review of the code so just wait for a green tick from Paul too.

PaulAllanSturm (approved):

PR #83, opened 2024-07-22
"Show Device Location History" sample updated

changanxian:
- Update the sample to follow the common design and reference other SDK with the same sample.


lib/samples/show_device_location_history/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+![Image of show device location history](show_device_location_history.png)
+
+## Use case
+
+You can track device location history and display it as lines and points on the map. The history can be used to visualize how the user moved through the world, to retrace their steps, or to create new feature geometry. An unmapped trail, for example, could be added to the map using this technique.
+
+## How to use the sample
+
+Tap 'Start tracking' to start tracking your location, which will appear as points on the map. A line will connect the points for easier visualization. Tap 'Stop tracking' to stop updating the location history. This sample uses a simulated data source to allow the sample to be useful on desktop/non-mobile devices. To track a user's real position, use the `LocationDataSource` instead.
```
PaulAllanSturm, line 13: The "T" in "Tracking" should be capitalized in 'Start tracking' and 'Stop tracking'

lib/samples/show_device_location_history/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## How it works
+
+
```
PaulAllanSturm, line 17: There are two blank lines here -- there should be only one.

lib/samples/show_device_location_history/README.md
```diff
@@ -0,0 +1,43 @@
...
+1. Create a graphics overlay to show each point and another graphics overlay for displaying the route line.
+2. Create a `SimulatedLocationDataSource` and initialize it with a polyline. Start the `SimulatedLocationDataSource` to begin receiving location updates.
+ - NOTE: To track a user's real position, use `LocationDataSource` instead.
+3. Subscribe to the `LocationChanged` event to handle location updates.
```
PaulAllanSturm, line 21: I think `LocationChanged` should be `onLocationChanged`

lib/samples/show_device_location_history/show_device_location_history_sample.dart
```diff
@@ -26,26 +26,39 @@ class ShowDeviceLocationHistorySample extends StatefulWidget {
...
 
   @override
-  State<ShowDeviceLocationHistorySample> createState() =>
-      _ShowDeviceLocationHistorySampleState();
+  State<ShowDeviceLocationHistorySample> createState() => Configure();
 }
 
-class _ShowDeviceLocationHistorySampleState
-    extends State<ShowDeviceLocationHistorySample> with SampleStateSupport {
+class Configure extends State<ShowDeviceLocationHistorySample>
```
PaulAllanSturm, line 32: I think we should leave the name as `_ShowDeviceLocationHistorySampleState` instead of `Configure`, so it's consistent with the other samples.

lib/samples/show_device_location_history/show_device_location_history_sample.dart
```diff
@@ -26,26 +26,39 @@ class ShowDeviceLocationHistorySample extends StatefulWidget {
...
+  // Create a controller for the map view.
   final _mapViewController = ArcGISMapView.createController();
-  double _latitude = 0.0;
-  double _longitude = 0.0;
-  double _heading = 0.0;
+  // A location data source to simulate location updates.
   final _locationDataSource = SimulatedLocationDataSource();
+  // Subscriptions to listen for location status changes.
   StreamSubscription? _statusSubscription;
+  // Subscriptions to listen for location changes.
```
PaulAllanSturm, line 40: "Subscriptions" should be "Subscription" here and the previous comment.

lib/samples/show_device_location_history/show_device_location_history_sample.dart
```diff
@@ -26,26 +26,39 @@ class ShowDeviceLocationHistorySample extends StatefulWidget {
...
+  // A flag for toggling location tracking.
+  var _enableTracking = false;
 
   @override
   void dispose() {
     _locationDataSource.stop();
     _locationSubscription?.cancel();
     _statusSubscription?.cancel();
+    _locationHistoryLineOverlay.graphics.clear();
+    _locationHistoryPointOverlay.graphics.clear();
```
PaulAllanSturm, line 61: I don't think there's a need to clear the graphics. They should be cleaned up when the garbage collector is triggered.

lib/samples/show_device_location_history/show_device_location_history_sample.dart
```diff
@@ -160,19 +180,44 @@ class _ShowDeviceLocationHistorySampleState
     });
```
PaulAllanSturm, line 180: I don't think we need `_statusSubscription` at all. There's nothing in the UI that gets updated when the status changes.

lib/samples/show_device_location_history/show_device_location_history_sample.dart
```diff
@@ -160,19 +180,44 @@ class _ShowDeviceLocationHistorySampleState
...
 
     try {
+      // Start the location data source.
       await _locationDataSource.start();
-      _locationSubscription = _locationDataSource.onLocationChanged
-          .listen(_handleLdsLocationChange);
+      // Listen for location changes.
+      _locationSubscription = _locationDataSource.onLocationChanged.listen(
+        _handleLdsLocationChange,
+      );
```
PaulAllanSturm, line 188: The call to `onLocationChanged.listen` should move after the try/catch block. It can't throw an ArcGISException, so there's no reason for it to be a part of the try/catch block.

lib/samples/show_device_location_history/show_device_location_history_sample.dart
changanxian: Here is my thought:

When a stateful widget (sample) is removed from the UI tree (Sample Viewer), it is not immediately destroyed by garbage collection (GC). Since the sample is retained by our Sample Viewer (_sampleWidget), it will persist until we switch to a new one or terminate the Sample Viewer. This results in unnecessary memory usage, so we need to clean it up as much as possible.

PaulAllanSturm: I wrote this up in Zenhub (issue 2023). We'll need to address it for all samples. There might be a more global solution -- perhaps the SampleListView can dispose of the sample when we navigate back.

PaulAllanSturm (approved):

PR #96, opened 2024-07-25
Add map image layer sample

shokhinaToj:
Created a new sample based on common design specifications. Added new README files and screenshot.


assets/generated_samples_list.json
```diff
@@ -880,14 +902,28 @@
   },
   "set_basemap": {
     "category": "Maps",
-    "description": "Set a map's basemap. A basemap is beneath all layers on a map and is used to provide visual reference for the operational layers.",
+    "description": "Change a map's basemap. A basemap is beneath all layers on a map and is used to provide visual reference for the operational layers.",
```
shokhinaToj, line 905: I think becasue I forgot to update this json after updating the metadata for this sample in the past, it update it now because I ran the script

PaulAllanSturm: Yes, it's fine to leave them. These changes showed up in other PRs too. Somebody will merge them first, and then everyone will have them and it'll straighten itself out.

lib/samples/add_map_image_layer/add_map_image_layer_sample.dart
```diff
@@ -0,0 +1,82 @@
...
+                ),
+              ),
+            ),
+          ],
+        ),
+      ),
+    );
+  }
+
+  void _onMapViewReady() async {
```
PaulAllanSturm, line 65: This should be named `onMapViewReady` instead of `_onMapViewReady`.

lib/samples/add_map_image_layer/add_map_image_layer_sample.dart
```diff
@@ -0,0 +1,82 @@
...
+class _AddMapImageLayerSampleState extends State<AddMapImageLayerSample>
+    with SampleStateSupport {
+  // Create a map view controller.
+  final _mapViewController = ArcGISMapView.createController();
+  // A flag for when the map view is ready.
+  var _ready = false;
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
```
PaulAllanSturm, line 37: Because of how simple this sample is, we really only need the Scaffold and the ArcGISMapView. We don't need the SafeArea, the Stack, or the _ready overlay thingy because there are no controls or tap handler. So those can all get deleted. And then there's no need for the _ready variable.

lib/samples/add_map_image_layer/add_map_image_layer_sample.dart
```diff
@@ -0,0 +1,82 @@
...
+  void _onMapViewReady() async {
+    // Create a map with a map image layer.
+    final map = ArcGISMap();
+    // Create a map image layer with a uri.
+    final mapImageLayer = ArcGISMapImageLayer.withUri(
+      Uri.parse(
+          'https://sampleserver5.arcgisonline.com/arcgis/rest/services/Elevation/WorldElevations/MapServer'),
+    );
+    // Set the map on the map view controller.
+    _mapViewController.arcGISMap = map;
```
PaulAllanSturm, line 74: I think in other samples we do this as the last step. That makes the tutorial read a little cleaner -- first create the map, then configure the map layers or whatever, then set the map on the controller.

PaulAllanSturm (approved):

assets/generated_samples_list.json
jenmerritt: This will be much smoother when we link up some of these scripts too and automate more, so it doesn't matter if it occasionally gets missed for now!

jenmerritt (approved):

PR #101, opened 2024-07-26
"Create Mobile Geodatabase" follow-up

changanxian:
- initially disable the two buttons
 - remove the feedback dialogs
 - update the button text


jenmerritt (approved):
Thanks @changanxian , this workflow is much cleaner 👍 

(Also apologies for the delay getting to this PR!)

PR #103, opened 2024-07-29
Set reference scale New Sample

shokhinaToj:
Created a new sample and added all the README files


lib/samples/set_reference_scale/set_reference_scale_sample.dart
```diff
@@ -0,0 +1,227 @@
...
+    setState(() => _ready = true);
+  }
+
+  StatefulBuilder showSettings(BuildContext newContext) {
+    // Show the settings dialog.
+    return StatefulBuilder(builder: (newContext, setNewState) {
+      return Dialog(
+        child: Padding(
+          padding: const EdgeInsets.all(8),
+          child: Column(
```
PaulAllanSturm, line 139: What happens when you rotate the device into landscape? This won't all fit. I think it needs a SingleChildScrollView around the whole thing.

lib/samples/set_reference_scale/set_reference_scale_sample.dart
```diff
@@ -0,0 +1,227 @@
...
+    }
+    _mapViewController.arcGISMap = _map;
+
+    // See if the map is ready.
+    setState(() => _ready = true);
+  }
+
+  StatefulBuilder showSettings(BuildContext newContext) {
+    // Show the settings dialog.
+    return StatefulBuilder(builder: (newContext, setNewState) {
```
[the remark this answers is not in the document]
shokhinaToj: I thought of giving it a diff name to distinguish it from the main build context. Is it unnecessary?

lib/samples/set_reference_scale/set_reference_scale_sample.dart
```diff
@@ -0,0 +1,227 @@
...
+                icon: const Icon(
+                  Icons.arrow_drop_down,
+                  color: Colors.deepPurple,
+                ),
+                elevation: 16,
+                style: const TextStyle(color: Colors.deepPurple),
+                // Set the onChanged callback to update the selected scale.
+                onChanged: (newScale) {
+                  updateScale(newScale!.toDouble());
+                },
```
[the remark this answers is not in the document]
shokhinaToj: I got rid of the updateScale() function but `newScale` is num so we would need to be casting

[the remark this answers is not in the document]
PaulAllanSturm: Maybe it needs casting, like `newScale as double`. But it should be smart enough to know it's a double.

Actually, I think I know what it is -- _referenceScaleList uses `double`, but _selectedScale uses `int`. That confuses the compiler and the best it can come up with is `num`. I bet if _selectedScale was a double, the compiler would figure out that newScale is supposed to be a double.

And actually it might make sense to get rid of _selectedScale, and just use _scale in its place. Yeah _selectedScale doesn't really do anything, and _scale is actually doing the work.

lib/samples/set_reference_scale/set_reference_scale_sample.dart
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  void updateScale(double newScale) {
+    // Update the scale.
+    setState(() => _scale = newScale);
+  }
+
+  void setScaleAndLayers() {
+    // Set the map scale to the reference scale and update the layers.
+    setState(() {
```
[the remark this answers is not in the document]
shokhinaToj: I just tested this and you are right! But thought since ArcGISMapView() is inside the widget tree, we would need to redraw it to update the scale and the layers?

lib/samples/set_reference_scale/set_reference_scale_sample.dart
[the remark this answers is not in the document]
shokhinaToj: Yeap you are right! I removed _selectedScale and it solved the problem with casting, great suggestion Paul!

lib/samples/set_reference_scale/set_reference_scale_sample.dart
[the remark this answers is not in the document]
PaulAllanSturm: We do our stuff via the _mapViewController, which kind of bypasses the setState/build mechanism.

Flutter does the same thing with some of its widgets, like the TextEditingController. Like you can call `_textEditingController.clear()` without having to call `setState()`. We adopted that idea for our _mapViewController. There's just so much stuff going on in a _mapViewController that we wouldn't want to build that all in to the widget tree.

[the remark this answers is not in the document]
shokhinaToj: That makes sense, thanks for the explanation!

lib/samples/set_reference_scale/set_reference_scale_sample.dart
```diff
@@ -0,0 +1,206 @@
...
+      );
+    });
+  }
+
+  void setScaleAndLayers() {
+    // Set the map scale to the reference scale.
+    _map.referenceScale = _scale;
+    // Update the layers to honor the reference scale.
+    for (final layer in _selectedFeatureLayers) {
+      var matchingLayers = _map.operationalLayers
```
PaulAllanSturm, line 200: Since `matchingLayers` will only be one thing, it should be `matchingLayer`. The plural kinda confused me.

lib/samples/set_reference_scale/set_reference_scale2.png
PaulAllanSturm, line 1: Do the images need to be updated after the change to the text styles?

lib/samples/set_reference_scale/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+Setting a reference scale on an `ArcGISMap` fixes the size of symbols and text to the desired height and width at that scale. As you zoom in and out, symbols and text will increase or decrease in size accordingly. When no reference scale is set, symbol and text sizes remain the same size relative to the `ArcGISMapView`.
+
+Map annotations are typically only relevant at certain scales. For instance, annotations to a map showing a construction site are only relevant at that construction site's scale. So, when the map is zoomed out that information shouldn't scale with the `ArcGISMapView`, but should instead remain scaled with the `ArcGISMap`.
+
+## How to use the sample
+
+* Use the drop-down menu through the Settings dialog to set the map's reference scale (1:500,000 1:250,000 1:100,000 1:50,000).
+* Tap the button to set the map scale to the reference scale.
+* Use the checkboxes in the Settings dialog to set which feature layers should honor the reference scale.
```
jenmerritt, line 18:
```suggestion
Tap the "Settings" button to load the settigs dialog. Use the drop-down menu to set the map's reference scale (1:500,000 1:250,000 1:100,000 1:50,000). You can choose which feature layers should honor the reference scale using the checkboxes. Tap the "Set to Reference Scale" button to set the map scale to the reference scale.
```

Appreciate you've copied the design with the steps here but I think the ordering needs changing to make things clearer, as the checkboxes are related to the reference scale, and the button setting the map scale is a different thing. I actually also think the design of the settings panel could also be slightly tweaked to help make this distinction between map and reference scale clearer. Right now it  also feels like I have to use the button to apply the settings.

Maybe have in this order:
- Settings
- A subheading "Reference Scale"
- Dropdown (as-is)
- Text "Apply Reference Scale to Layers"
- Checkboxes (as is)
- A subheading "Map Scale"
- Display the current map scale (see other SDK implementations)
- Button, change text to "Set to Reference Scale"

lib/samples/set_reference_scale/set_reference_scale_sample.dart
```diff
@@ -0,0 +1,206 @@
...
+                  // Set the selected scale
+                  value: _scale,
+                  icon: const Icon(
+                    Icons.arrow_drop_down,
+                    color: Colors.deepPurple,
+                  ),
+                  // Set the onChanged callback to update the selected scale.
+                  onChanged: (newScale) {
+                    setState(() => _scale = newScale!);
+                  },
```
jenmerritt, line 152: I'm currently only seeing the reference scale value updated when a checkbox is selected/unselected, I don't see it update when I use the dropdown. Feels like there is something not quite right with the state update here.

lib/samples/set_reference_scale/set_reference_scale_sample.dart
```diff
@@ -0,0 +1,206 @@
...
+              mainAxisAlignment: MainAxisAlignment.spaceAround,
+              children: [
+                Center(
+                  child: Text('Settings',
+                      style: Theme.of(context).textTheme.headlineMedium),
+                ),
+                // Add a dropdown button for setting a new reference scale.
+                DropdownButton(
+                  alignment: Alignment.center,
+                  hint: const Text('Select a New Reference Scale'),
```
jenmerritt, line 142:
```suggestion
```
I don't think we need this, since we start with a default value set.

lib/samples/set_reference_scale/set_reference_scale_sample.dart
```diff
@@ -0,0 +1,206 @@
...
+                      ),
+                  ],
+                ),
+                // Add a button to set the map scale to the reference scale and update the layers.
+                ElevatedButton(
+                  onPressed: () {
+                    setScaleAndLayers();
+                    Navigator.pop(context);
+                  },
+                  child: const Text('Set Map Scale to Reference Scale'),
```
jenmerritt, line 185: This is also sort of related to my comment above in the readme. The button here should just "set the map scale to the reference scale". 

The layer reference scales should be updated when you toggle the checkboxes and/or update the reference scale in the dropdown. They are 2 distinct events. So I think you'll need to separate this method out.

jenmerritt (commented):
The main thing from my comments below is that the action of updating the reference scale and the action of setting the map scale to the reference scale are 2 distinct actions.

shokhinaToj:
@jenmerritt @PaulAllanSturm I updated the code based on Jen's suggestions (sorry I think I was little bit confused about the order of things). But now it makes more sense! I followed what iOS team did for this sample - could you please verify my work again :)

lib/samples/set_reference_scale/set_reference_scale_sample.dart
shokhinaToj: Ohh gotcha, will update this!

lib/samples/set_reference_scale/README.md
shokhinaToj: Yes this is what I initially saw for iOS but thought since the common design spec doesn't mention these, maybe they are unnecessary. I will make sure to update the order but should we suggest this change to the common design readme?

lib/samples/set_reference_scale/set_reference_scale_sample.dart
```diff
@@ -0,0 +1,225 @@
...
+              ],
+            ),
+          ),
+        ),
+      );
+    });
+  }
+
+  // Create a function to format the scale.
+  String Function(Match) mathFunc = (Match match) => '${match[1]},';
```
PaulAllanSturm, line 224: `mathFunc` should be `matchFunc`, although to make it more descriptive it could maybe be `matchFormatter`. And `reg` could also use a more descriptive name, such as `digitGroupRegex` or something like that.

lib/samples/set_reference_scale/set_reference_scale_sample.dart
```diff
@@ -0,0 +1,225 @@
...
+  // Create a flag for when the map view is ready and controls can be used.
+  var _ready = false;
+  // Create a regular expression to format the scale.
+  final reg = RegExp(r'(\d{1,3})(?=(\d{3})+(?!\d))');
+
+  @override
+  void initState() {
+    super.initState();
+    // Add scales to the list.
+    _referenceScaleList.addAll([
```
PaulAllanSturm, line 52: Now that you have a mechanism to do `replaceAllMapped(reg, mathFunc)`, maybe that can be used here, for simplicity and consistency. Something like:
```dart
for (final value in [500000.0, 250000.0, 100000.0, 50000.0]) {
  _referenceScaleList.add(const DropdownMenuItem(
      value: value,
      child: Text('1:${value.toInt().toString().replaceAllMapped(reg, mathFunc)}'),
    ),
  );
}
```
And you could maybe make a helper function that does the whole `'1:${value.toInt().toString().replaceAllMapped(reg, mathFunc)}'` thing. Regexes are notoriously hard to understand so it might be nice to bundle it up into a helper function with a friendly comment. Maybe like `String formatAsScale(double value) { ... }`.

shokhinaToj: yes this would make sense!

PaulAllanSturm (approved):

lib/samples/set_reference_scale/README.md
jenmerritt: It's probably not necessary in this case, because the UI is so different for desktop vs mobile. In this case we're needing to make everything quite linear because it's in a popup type dialog, whereas desktop they'd have all the controls on the screen.

lib/models/samples_widget_list.dart
```diff
@@ -81,6 +82,7 @@ const sampleWidgets = <String, Widget>{
   'search_with_geocode': SearchWithGeocodeSample(),
   'select_features_in_feature_layer': SelectFeaturesInFeatureLayerSample(),
   'set_basemap': SetBasemapSample(),
+  'set_reference_scale': SetReferenceScaleSample(),
```
jenmerritt, line 85: v.next has just been updated with changes to how this class is defined. See if you can update your branch and match the new format here.

lib/samples/set_reference_scale/set_reference_scale_sample.dart
```diff
@@ -0,0 +1,222 @@
...
+}
+
+class _SetReferenceScaleState extends State<SetReferenceScaleSample>
+    with SampleStateSupport {
+  // Create a controller for the map view.
+  final _mapViewController = ArcGISMapView.createController();
+  // Create a list of dropdown menu items to load all the reference scales.
+  final _referenceScaleList = <DropdownMenuItem<double>>[];
+  // Create a list of selected feature layers.
+  final _selectedFeatureLayers = <String>[];
```
jenmerritt, line 36: This is looking good! The only thing that jumps out now is that the reference scale is being updated in all layers when I toggle initially despite them being all unchecked- so I think maybe we need to check the initial state. Maybe they can all start as selected - so we just need to ensure the checkboxes reflect this. I haven't had chance to look in detail at why, but maybe have a play around. Once I toggle the checkboxes on/off, it works as expected.

lib/models/samples_widget_list.dart
shokhinaToj: Ok will do!

lib/samples/set_reference_scale/set_reference_scale_sample.dart
shokhinaToj: Oh yes I think I know what you mean - I noticed they are all checked out for iOS sample so I will make sure to update that in onMapViewReady and have them display the initial state of the layer.refenceSymbol

jenmerritt (approved):
👍

PR #106, opened 2024-07-31
Remove "sample" from Widget names and filenames for each sampleShokhina toj/1843 remove sample keyword

shokhinaToj:
Removed "sample" from source files and widgets. Adjusted the metadata references (snippets) and widget list too.


PaulAllanSturm (approved):
Probably we need a separate issue to update flutter/docs/Samples.md (it mentions _sample.dart and Sample class name in at least one place).

PR #107, opened 2024-07-31
Clear some warnings

PaulAllanSturm:
`flutter analyze` picked up a few things


jenmerritt (approved):

jenmerritt:
Potentially could be good to see if there are any candidates like this worth cherry picking into `main` after we merge from the release tag...

PR #111, opened 2024-08-02
Apply new Linter Rules

shokhinaToj:
Applied most of the linter rules from here https://github.com/VeryGoodOpenSource/very_good_analysis/blob/main/lib/analysis_options.5.1.0.yaml
The following linter rules were removed due to requiring fixes in many places and restructuring our code:
    - always_declare_return_types
    - always_put_required_named_parameters_first
    - always_use_package_imports 
    - avoid_dynamic_calls 
    - avoid_redundant_argument_values 
    - avoid_void_async 
    - cascade_invocations
    - cast_nullable_to_non_nullable 
    - depend_on_referenced_packages 
    - lines_longer_than_80_chars  
    - prefer_final_locals 
    - prefer_if_elements_to_conditional_expressions 
    - prefer_int_literals 
    - public_member_api_docs  
    - sort_constructors_first 
    - type_annotate_public_apis 
    - unawaited_futures 
    - unnecessary_lambdas 
    - use_colored_box 
    - use_setters_to_change_properties 
    - avoid_positional_boolean_parameters
    - sort_pub_dependencies 
    - unnecessary_statements

If we need any of these, I can go back and add it!


lib/samples/add_feature_collection_layer_from_table/add_feature_collection_layer_from_table.dart
```diff
@@ -116,7 +116,7 @@ class _AddFeatureCollectionLayerFromTableState
     // Create a feature collection table for the geometry type with a list of fields and a spatial reference.
     final polylineTable = FeatureCollectionTable(
       fields: [
-        Field.text(name: 'Boundary', alias: 'Boundary Name', length: 50)
+        Field.text(name: 'Boundary', alias: 'Boundary Name', length: 50),
```
jenmerritt, line 119: This comma rule (in multiple places) goes against our code standards so we either need to adopt it/update those, or not have the linter. Will let others weigh in, I don't really mind.

PaulAllanSturm: I much prefer an automated linter rule, even if it's not exactly what we were doing manually. We can prune it from our coding standard.

lib/samples/apply_scheduled_updates_to_preplanned_map_area/apply_scheduled_updates_to_preplanned_map_area.dart
```diff
@@ -46,8 +46,8 @@ class _ApplyScheduledUpdatesToPreplannedMapAreaState
   // The Active mobile map package.
   MobileMapPackage? _mobileMapPackage;
   // Offline task and parameters used for updating the map package.
-  OfflineMapSyncTask? _offlineMapSyncTask;
-  OfflineMapSyncParameters? _mapSyncParameters;
+  late OfflineMapSyncTask? _offlineMapSyncTask;
+  late OfflineMapSyncParameters? _mapSyncParameters;
```
PaulAllanSturm, line 50: This change looks weird to me -- as far as I know, we should be able to declare optionals without `late`, because they default to `null`. I don't see what's special about these two, compared to `_mobileMapPackage` just above.

Which linter rule drove this change? (Like, if you change it back, what is the analyzer message?)

lib/samples/add_feature_layer_with_time_offset/add_feature_layer_with_time_offset.dart
```diff
@@ -187,7 +187,7 @@ class _AddFeatureLayerWithTimeOffsetState
 
     // The URL of the feature layer showing hurricanes.
     final featureLayerUri = Uri.parse(
-        'https://sampleserver6.arcgisonline.com/arcgis/rest/services/Hurricanes/MapServer/0');
+        'https://sampleserver6.arcgisonline.com/arcgis/rest/services/Hurricanes/MapServer/0',);
```
PaulAllanSturm, line 190: With the addition of the new commas in all the files, the code should end up getting reformatted a little, the next time anyone does `dart format`. The automated `dart fix` stuff doesn't do that for us, so we should run `dart format` on the sample repo. Could you run `dart format` in the samples repo? That'll clean up a bunch of stuff. (Unfortunately, it'll also clean up some stuff caused by the recent "remove Samples from the class name" change, but that should be fine.)

lib/samples/add_feature_collection_layer_from_table/add_feature_collection_layer_from_table.dart
PaulAllanSturm: Here's the linter rule explanation: https://dart.dev/tools/linter-rules/require_trailing_commas
It's going to recommend more commas than what we were doing, but I think that's fine. It has some reasonable exceptions built-in.

lib/samples/apply_scheduled_updates_to_preplanned_map_area/apply_scheduled_updates_to_preplanned_map_area.dart
shokhinaToj: This is what the linter says ` Use 'late' for private members with a non-nullable type. Try making adding the modifier 'late'. • use_late_for_private_fields_and_variables`
But what you said about it being inconsistent makes sens becase this linter rule is marked "[experimental](https://dart.dev/tools/linter-rules/use_late_for_private_fields_and_variables)" so we might need to use this one with caution (or remove it?)

PaulAllanSturm: Actually I think I see what it's saying. Every time we access `_offlineMapSyncTask`, we use `!`, which means we're treating it as if it is never null. So there's no point in us declaring it nullable. It makes sense that it's telling us to use `late` in that case. But it's a little off -- the underlying problem is really that we shouldn't have used `?` in the declaration. So I think we should have this:
```dart
  late OfflineMapSyncTask _offlineMapSyncTask;
  late OfflineMapSyncParameters _mapSyncParameters;
```
This change will mean we have to remove the `!` wherever we access the variable, which is reasonable. Then once those changes are made, all the lint rules should be satisfied.

shokhinaToj: okay, I will make sure to update them!

shokhinaToj:
Spoke to Paul about using the exisitng linter packages provided in pub.dev - we concluded that it won't hurt to have stricter rules customized by us and our users won't be affected by them since they would have their own rules for their projects

PaulAllanSturm (approved):

PR #114, opened 2024-08-06
Fixed a typo in a new sample

shokhinaToj:
Missing 'n' in 'settings' in README file

### DIFF
--- a/lib/samples/set_reference_scale/README.md
+++ b/lib/samples/set_reference_scale/README.md
@@ -13,7 +13,7 @@ Map annotations are typically only relevant at certain scales. For instance, ann
 
 ## How to use the sample
 
-Tap the "Settings" button to load the settigs dialog. Use the drop-down menu to set the map's reference scale (1:500,000 1:250,000 1:100,000 1:50,000). You can choose which feature layers should honor the reference scale using the checkboxes. Tap the "Set to Reference Scale" button to set the map scale to the reference scale.
+Tap the "Settings" button to load the settings dialog. Use the drop-down menu to set the map's reference scale (1:500,000 1:250,000 1:100,000 1:50,000). You can choose which feature layers should honor the reference scale using the checkboxes. Tap the "Set to Reference Scale" button to set the map scale to the reference scale.
 
 ## How it works
 


JamesMBallard (approved):

lib/samples/set_reference_scale/set_reference_scale.dart
```diff
@@ -124,100 +124,140 @@ class _SetReferenceScaleState extends State<SetReferenceScale>
...
-                        value: _selectedFeatureLayers.contains(layer),
-                        onChanged: (value) {
-                          setNewState(() {
-                            // Update the selected feature layers list.
-                            if (value ?? false) {
-                              _selectedFeatureLayers.add(layer);
-                            } else {
-                              _selectedFeatureLayers.remove(layer);
-                            }
+  Widget showSettings(BuildContext context) {
```
PaulAllanSturm, line 127: Elsewhere we name this method "buildSettings", to make it clear that it's related to the usual "build" method.

lib/samples/set_reference_scale/set_reference_scale.dart
```diff
@@ -124,100 +124,140 @@ class _SetReferenceScaleState extends State<SetReferenceScale>
...
+    return Container(
+      padding: const EdgeInsets.fromLTRB(
+        20.0,
+        20.0,
+        20.0,
+        0,
```
PaulAllanSturm, line 133: Instead of `0` at the bottom, elsewhere we have:
```dart
        max(
          20.0,
          View.of(context).viewPadding.bottom /
              View.of(context).devicePixelRatio,
        ),
```
This is unfortunately necessary for devices that do not have safe-area controls on the bottom, such as iPhones with hardware "home" buttons (e.g. iPhone SE 3rd gen).

lib/samples/set_reference_scale/set_reference_scale.dart
```diff
@@ -124,100 +124,140 @@ class _SetReferenceScaleState extends State<SetReferenceScale>
...
+                      Text('Map Scale',
+                          style: Theme.of(context).textTheme.titleMedium),
+                      const Spacer(),
+                      Text(formatAsScale(_mapViewController.scale),
+                          style: Theme.of(context).textTheme.titleSmall),
+                    ],
+                  ),
+                  // Add a button to set the map scale to the reference scale.
+                  Padding(
+                    padding: const EdgeInsets.only(bottom: 20.0),
```
PaulAllanSturm, line 243: Oh I see where you put the bottom padding. Unfortunately we need the "max" formula from above to cover devices both with and without safe-area controls.

lib/samples/set_reference_scale/set_reference_scale.dart
shokhinaToj: Got it!

lib/samples/set_reference_scale/set_reference_scale.dart
shokhinaToj: Yeah the reason I put it there is because it was cutting off UI at bottom in an ugly way so this bottom cuts it only when u scroll all the way to the button - but your hardware limitation explanation makes sense!

PaulAllanSturm (approved):

PR #132, opened 2024-08-27
New Sample: Add WMS Layer

jenmerritt:
Add WMS Layer Sample:
 - Adds all required sample files per the design
 - Updated generated sample list


lib/samples/add_wms_layer/add_wms_layer.dart
```diff
@@ -0,0 +1,65 @@
...
+import 'package:flutter/material.dart';
+
+class AddWmsLayer extends StatefulWidget {
+  const AddWmsLayer({super.key});
+
+  @override
+  State<AddWmsLayer> createState() => _AddWmsLayerState();
+}
+
+class _AddWmsLayerState extends State<AddWmsLayer> {
```
PaulAllanSturm, line 26: Should we always do `with SampleStateSupport` even for samples that don't call setState? I think so, because the danger is we'll forget it when we need it, but not get any warning.

lib/samples/add_wms_layer/add_wms_layer.dart
```diff
@@ -0,0 +1,65 @@
...
+    // Load the layer and get its extent.
+    await wmsLayer.load();
+    final layerExtent = wmsLayer.fullExtent;
+    // Create a map with a basemap style.
+    final map = ArcGISMap.withBasemapStyle(BasemapStyle.arcGISLightGray);
+    // Add the WMS layer to the maps operational layers.
+    map.operationalLayers.add(wmsLayer);
+    // Set the map to the map view controller.
+    _mapViewController.arcGISMap = map;
+    // Set the viewpoint to the layers extent.
```
PaulAllanSturm, line 60: "layers" should be "layer's"

PaulAllanSturm: I would move the line `final layerExtent = wmsLayer.fullExtent;` down here. Otherwise the thought is broken up in two places. (And then adjust the comment above the `load()` call.)

lib/samples/add_wms_layer/add_wms_layer.dart
jenmerritt: Yeah fair enough, good spot. This wasn't deliberate. I use the `stfl` shortcut to create widgets and just forgot about it.

kmueller-gis (approved):

lib/samples/add_wms_layer/add_wms_layer.dart
jenmerritt: I'm going to also move the load down, as the reason we have to explicility load is in order to get the extent so want to keep that comment together.

PaulAllanSturm (approved):

PR #143, opened 2024-09-09
New sample: Identify features in wms layer

jenmerritt:
Adds sample as per design.
Requires additional package which I'll add to our docs.
We could enhance this sample by using Callout, so I'll also note that for future ref.


lib/samples/identify_features_in_wms_layer/README.md
```diff
@@ -0,0 +1,40 @@
...
+2. When the map is tapped, `ArcGISMapViewController.identifyLayer` is used to find matching results within the WMS layer.
+3. If there is a matching feature, the HTML property is taken via a lookup in the feature's attribute dictionary.
+4. This particular server will produce an identify result with an empty table when there is no identified feature. In all other cases, a table with an **OBJECTID** column is added. This sample checks for the presence of **OBJECTID** in the HTML, and doesn't display the result if it is missing.
+5. The HTML is displayed in an alert dialog.
+
+**Note**: the service returns HTML regardless of whether there was an identify result. The sample uses a simple rule to hide empty results.
+
+## Relevant API
+
+* ArcGISMapViewController.identifyLayerAsync
```
PaulAllanSturm, line 27: "identifyLayerAsync" should be "identifyLayer"

PaulAllanSturm: (and regenerate metadata & generated_samples_list.json)

lib/samples/identify_features_in_wms_layer/identify_features_in_wms_layer.dart
```diff
@@ -0,0 +1,191 @@
...
+                  ),
+                ),
+              ],
+            ),
+            // Display a banner with instructions.
+            SafeArea(
+              child: IgnorePointer(
+                child: Container(
+                  padding: const EdgeInsets.all(10),
+                  color: Colors.white.withOpacity(0.7),
```
PaulAllanSturm, line 64: download_preplanned_map_area.dart and identify_layer_features.dart have white-on-black. Should these banner styles be consistent one way or the other?

lib/samples/identify_features_in_wms_layer/identify_features_in_wms_layer.dart
```diff
@@ -0,0 +1,191 @@
...
+    // Show a dialog containing a web view widget.
+    showDialog(
+      context: context,
+      builder: (context) => AlertDialog(
+        title: Text(
+          'Identify Result',
+          style: Theme.of(context).textTheme.titleMedium,
+        ),
+        content: SizedBox(
+          height: 150,
```
PaulAllanSturm, line 163: does this work in landscape orientation? Otherwise we might need like a BoxConstraints/maxHeight/SingleChildScrollView sort of thing.

kmueller-gis (approved):

lib/samples/identify_features_in_wms_layer/README.md
jenmerritt: Good catch

lib/samples/identify_features_in_wms_layer/identify_features_in_wms_layer.dart
jenmerritt: I flipped it since the Basemap is dark as otherwise it's not very visible. So I figured either one is fine, since they're consistent just reversed.

lib/samples/identify_features_in_wms_layer/identify_features_in_wms_layer.dart
jenmerritt: Yeah I tested quite a bit in landscape as actually it's better viewed in landscape as the whole HTML table width fits. The HTML content height is consistent so it looks the same height wise.

assets/generated_samples_list.json
```diff
@@ -981,6 +981,39 @@
...
+    "description": "Identify features in a WMS layer and display the associated popup content.",
+    "ignore": false,
+    "images": [
+      "identify_features_in_wms_layer.png"
+    ],
+    "keywords": [
+      "OGC",
+      "WMS",
+      "web map service",
+      "ArcGISMapViewController.identifyLayerAsync",
```
PaulAllanSturm, line 995: It still shows up here

PaulAllanSturm (approved):

PR #157, opened 2024-09-25
New sample: Snap Geometry Edits

jenmerritt:
Adds the snap geometry edits sample. 

One thought is that I've re-used the String extension. Are we happy with extensions in individual samples or do we want to start a util class when it's reused?


PaulAllanSturm:
> One thought is that I've re-used the String extension. Are we happy with extensions in individual samples or do we want to start a util class when it's reused?

Yes we should get to this, now or later. We already have SampleStateSupport as well.

I think we should do it as an internal package/library, so we have just one import statement, like 
```dart
import 'package:arcgis_maps_sdk_flutter_samples/sample_support.dart';
```
(I don't like having a misc bucket called "utils".)

We should also have some policy about what we put in there. It should only be things that aren't central to the tutorial flow. Anything tied directly to the tutorial should be spelled out inline, even if that means a bit of duplication. But other stuff we could/should stash in the support package, things like the overlay !_ready widget and a generic showAlertDialog method. And maybe the "Settings" bottomSheet widget. Especially stuff that is bulky and crowds out the tutorial content.

The downside is that users can't do a complete cut-n-paste of the sample code and have it work out-of-the-box. But I think we have to accept that tradeoff, especially as we add more complexity to the sample app.

lib/samples/snap_geometry_edits/snap_geometry_edits.dart
```diff
@@ -0,0 +1,679 @@
...
+          iconDisabledColor: Colors.grey,
+          style: const TextStyle(color: Colors.deepPurple),
+          value: _selectedGeometryType,
+          items: _geometryTypeMenuItems,
+          // If the geometry editor is already started then we don't enable editing with another geometry type.
+          onChanged: _geometryEditorIsStarted
+              ? null
+              : (geometryType) => geometryType == null
+                  ? null
+                  : startEditingWithGeometryType(geometryType as GeometryType),
```
PaulAllanSturm, line 315: `geometryType as GeometryType` should be `geometryType!` since all we're doing is getting rid of the nullability.

But I don't really like the `? null` on line 314. We're not really returning a value here. I think what we're trying to express is just
```dart
(geometryType) => if (geometryType != null) startEditingWithGeometryType(geometryType!),
```

kmueller-gis:
When I ran the sample, snapping is off by default. Since this is the snapping sample, shouldn't that be on by default?

kmueller-gis (commented):
When I ran the sample, snapping is off by default. Since this is the snapping sample, shouldn't that be on by default?

lib/samples/snap_geometry_edits/snap_geometry_edits.dart
```diff
@@ -0,0 +1,679 @@
...
+        ),
+        Row(
+          mainAxisAlignment: MainAxisAlignment.center,
+          children: [
+            const Text('Enable all'),
+            // A checkbox to enable all source settings in the category.
+            Checkbox(
+              value: allSourceSettings.firstWhereOrNull(
+                    (snapSourceSettings) =>
+                        snapSourceSettings.isEnabled == false,
```
PaulAllanSturm, line 624: `== false` is kinda weird. This could just be `!snapSourceSettings.isEnabled`

lib/samples/snap_geometry_edits/snap_geometry_edits.dart
```diff
@@ -0,0 +1,679 @@
...
+        ),
+        const SizedBox(height: 20),
+      ],
+    );
+  }
+
+  List<DropdownMenuItem<GeometryType>> configureGeometryTypeMenuItems() {
+    // Returns a list of drop down menu items for each geometry type.
+    return _geometryTypes.map((type) {
+      return DropdownMenuItem(
```
PaulAllanSturm, line 650: I think I would use the `=>` notation here instead of `{ ... }` since it's only executing a single statement. Removes the need for the `return`.

lib/samples/snap_geometry_edits/snap_geometry_edits.dart
jenmerritt: Yeah you're right, I often have things more verbose when I'm figuring things out and then refactor. Missed this one, will remove.

lib/samples/snap_geometry_edits/snap_geometry_edits.dart
jenmerritt: I'd tried simpler statements here in the other sample but couldn't get there. The UI dropdown doesn't properly get disabled. I'll have another go at simplifying.

lib/samples/snap_geometry_edits/snap_geometry_edits.dart
```diff
@@ -0,0 +1,679 @@
...
+    final graphics = identifyResult.graphics;
+    if (graphics.isNotEmpty) {
+      final graphic = graphics.first;
+      if (graphic.geometry != null) {
+        final geometry = graphic.geometry!;
+        // Hide the selected graphic so that only the version of the graphic that is being edited is visible.
+        graphic.isVisible = false;
+        // Set the graphic as the selected graphic and also set the selected geometry type to update the UI.
+        setState(() {
+          _selectedGraphic = graphic;
```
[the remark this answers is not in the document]
jenmerritt: Hmm I think you're right

lib/samples/snap_geometry_edits/snap_geometry_edits.dart
jenmerritt: The actual callback needs to be null for it to be properly disabled, (i.e. greyed out and not able to interact with, as opposed to just can't select another option which is a bad user experience). So if I check `_geometryEditorIsStarted` I can properly set the DropDownButton to null if it is. Note that then the `geometryType` is an `Object?` hence the typecast there.

I've made a tweak to the existing code. The other option would be the below but the UI is a bit weird. I could style the individual items to look more un-selectable and slightly improve things.

```
          onChanged: (geometryType) {
            if (!_geometryEditorIsStarted && geometryType != null) {
              startEditingWithGeometryType(geometryType);
            }
          },
```

PaulAllanSturm: Ah ok, I think I see what's going on. This is the first time we're using `null` as the value of `onChanged` to disable the widget. And that's correct, so don't change that.

Unfortunately that introduces a small oddity -- because of that `null`, the compiler infers that the widget is a `DropdownButton<Object>` rather than a `DropdownButton<GeometryType>`. The compiler can't be any more specific about the `null`, so it falls back to `<Object>`. And because of that, you end up with `geometryType` being an `Object?` rather than a `GeometryType`.

I'm not sure the best way to address that -- probably if you changed `DropdownButton` to `DropdownButton<GeometryType>`, that would force the issue and everything would go back to normal. But I don't like that as a general rule. Maybe if you change `? (geometryType) {` ... to ... `? (GeometryType? geometryType) {`, that might give the compiler what it needs to narrow down the type.

Or maybe there's a way to cast the `null` to a `ValueChanged<GeometryType?>?`. I wonder if you just do `null as ValueChanged<GeometryType?>?`.

kmueller-gis (approved):

lib/samples/snap_geometry_edits/snap_geometry_edits.dart
jenmerritt: Interesting! I've gone with `(GeometryType? geometryType)`.  `DropdownButton<GeometryType>` does also work.

jenmerritt:
Ready for re-review @PaulAllanSturm , thanks!

PaulAllanSturm (approved):

PR #159, opened 2024-09-25
Upgrade to Flutter 3.24.3

PaulAllanSturm:
Changes to keep up with the upgrade to Flutter 4.24.3


jenmerritt (approved):
I've also matched and tested these on the new tutorial solution too that Steve is working on.

PR #163, opened 2024-10-02
New sample: Manage Features

jenmerritt:
Adds Manage Features sample per the design.
Tested on iOS simulator and Android device.


lib/samples/manage_features/manage_features.dart
```diff
@@ -0,0 +1,433 @@
...
+            DropdownMenuItem(
+              onTap: () => setState(
+                () => _selectedOperation == operation,
+              ),
+              value: operation,
+              child: Text(getLabel(operation)),
+            ),
+          ),
+        )
+        .toList();
```
PaulAllanSturm, line 70: I'm not sure if it matters, but I would flip this around a bit like so:
```dart
_featureManagementOptions.addAll(
  FeatureManagementOperation.values.map(
    (operation) => DropDownMenuItem(
      // ...
    ),
  ),
);
```
That way it reads a bit smoother to me and it saves you the trouble of having to call `.toList()`. It's a bit obscure that you have to call `.toList()` in order to make the calls to `_featureManagementOptions.add()` happen.

lib/samples/manage_features/manage_features.dart
```diff
@@ -0,0 +1,433 @@
...
+            ),
+          ],
+        ),
+      ),
+    );
+  }
+
+  void onMapViewReady() async {
+    // Create and load a service geodatabase from a service URL.
+    const featureServiceURL =
```
PaulAllanSturm, line 143: I think elsewhere we'd name URL values with "Uri" at the end, so `featureServiceUri`

lib/samples/manage_features/manage_features.dart
```diff
@@ -0,0 +1,433 @@
...
+    final table = serviceGeodatabase.getTable(layerId: 0);
+    if (table != null) {
+      _damageServiceFeatureTable = table;
+      // Load the table.
+      await _damageServiceFeatureTable.load();
+      // Get the required field from the table - in this case, damage type.
+      final damageTypeField =
+          table.fields.firstWhere((field) => field.name == damageTypeFieldName);
+      // Get the domain for the field.
+      final domain = damageTypeField.domain as CodedValueDomain;
```
PaulAllanSturm, line 160: I don't want to add unnecessary complication, but I think the `as CodedValueDomain` hides a sneaky null dereference, given that the `domain` property is nullable. At least we should be explicit about it with `damageTypeField.domain! as CodedValueDomain`.

Otherwise we should maybe decide how picky the samples should be in making sure that the service continues to be configured as originally designed. Like the `firstWhere` will throw a `StateError` if the properly-named field can't be found, and the `load()` will throw if the URL goes bad.

lib/samples/manage_features/manage_features.dart
```diff
@@ -0,0 +1,433 @@
...
+    } else {
+      showMessageDialog('Unable to access the required feature table.');
+    }
+  }
+
+  void onTap(Offset localPosition) async {
+    // Configure actions when a user taps on the map, depending on the selected operation.
+    if (_selectedOperation == FeatureManagementOperation.create) {
+      // Create a feature if create is selected.
+      createFeature(localPosition);
```
PaulAllanSturm, line 195: I think we want `onTap` to `await` on each of the three possibilities (createFeature, updateGeometry, and identifyAndSelectFeature). Which means each of those would have to be declared to return `Future<void>` I think.

lib/samples/manage_features/manage_features.dart
```diff
@@ -0,0 +1,433 @@
...
+    // Get the normalized geometry for the tapped location and use it as the feature's geometry.
+    final geometry = _mapViewController.screenToLocation(screen: localPosition);
+    if (geometry != null) {
+      final normalizedGeometry =
+          GeometryEngine.normalizeCentralMeridian(geometry);
+      feature.geometry = normalizedGeometry;
+
+      // Set feature attributes.
+      feature.attributes['typdamage'] = 'Minor';
+      feature.attributes['primcause'] = 'Earthquake';
```
PaulAllanSturm, line 222: We have `damageTypeFieldName` we can use instead of `'typdamage'`, but we don't have a variable for `'primcause'`.

I'm generally of the opinion that defining a member variable like this is more trouble than it's worth. Like in this case if we needed to change `damageTypeFieldName` for some reason, we'd still have to do a global search-and-replace for errant hardcoded uses of `'typdamage'` just to be sure. We could add a member variable for `'primcause'`, and then add one for `'objectid'`, but that seems weird. And it also seems weird to have an exception just for `'objectid'`.

lib/samples/manage_features/manage_features.dart
```diff
@@ -0,0 +1,433 @@
...
+            DropdownMenuItem(
+              onTap: () => setState(
+                () => _selectedDamageType == value.name,
+              ),
+              value: value.name,
+              child: Text(value.name),
+            ),
+          ),
+        )
+        .toList();
```
PaulAllanSturm, line 349: same as above about flipping it around

lib/samples/manage_features/manage_features.dart
jenmerritt: I put it all in a try catch since i was already displaying an error message in one case.

lib/samples/manage_features/manage_features.dart
jenmerritt: Ok, I'd removed those as wasn't sure if we were waiting but I wasn't 100% either way on that so I've added back in.

lib/samples/manage_features/manage_features.dart
jenmerritt: Removed the variable

lib/samples/manage_features/manage_features.dart
```diff
@@ -55,19 +55,19 @@ class _ManageFeaturesState extends State<ManageFeatures> {
...
                 () => _selectedOperation == operation,
               ),
               value: operation,
               child: Text(getLabel(operation)),
             ),
-          ),
-        )
-        .toList();
+          )
+          .toList(),
```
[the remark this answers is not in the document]
jenmerritt: You're right it is redundant 👍

PR #166, opened 2024-10-03
Geometry Editor Sample updates

jenmerritt:
I shared our sample viewer with Shelly to check out our Geometry Editor samples in particular. She had a couple of thoughts and these may feed back to explicit instructions in the design from that team, but not as a high priority.

This PR:
- shows the editing toolbar by default
- enables all the snap sources by default (this is where we'll differ from the design, but it is to be discussed separately) - setState required I think due to the checkbox updates.
- moves the enable all checkboxes to the top of each section - this was actually a divergence I hadn't spotted.

No rush on this PR, just made the changes while it was fresh.


lib/samples/snap_geometry_edits/snap_geometry_edits.dart
```diff
@@ -574,9 +586,38 @@ class _SnapGeometryEditsState extends State<SnapGeometryEdits>
...
+            ),
+            Row(
+              children: [
+                const Text('Enable all'),
+                // A checkbox to enable all source settings in the category.
+                Checkbox(
+                  value: allSourceSettings.firstWhereOrNull(
+                        (snapSourceSettings) => !snapSourceSettings.isEnabled,
+                      ) ==
+                      null,
```
PaulAllanSturm, line 604: I'm probably a little late, but this could be
```dart
value: allSourceSettings.every((snapSourceSettings) => snapSourceSettings.isEnabled),
```

PaulAllanSturm: Which saves us the trouble of having `collection` as a dependency, I think

lib/samples/snap_geometry_edits/snap_geometry_edits.dart
```diff
@@ -574,9 +586,38 @@ class _SnapGeometryEditsState extends State<SnapGeometryEdits>
...
+                  onChanged: (allEnabled) {
+                    if (allEnabled != null) {
+                      allSourceSettings
+                          .map(
+                            (snapSourceSettings) =>
+                                snapSourceSettings.isEnabled = allEnabled,
+                          )
+                          .toList();
+                      // We must redraw to update the checkbox UI.
+                      setState(() {});
```
PaulAllanSturm, line 614: I think, at least for consistency, the `setState` should be wrapped around the `snapSourceSettings.isEnabled = allEnabled` a couple lines up.

lib/samples/snap_geometry_edits/snap_geometry_edits.dart
jenmerritt: Oh nice. I'll pick this back up in a couple of weeks so I can double check and update the dependency list if we can lose it.

jenmerritt: That worked - flipping the logic! Thanks @PaulAllanSturm , I'll also create a PR to our dependency list.

lib/samples/snap_geometry_edits/snap_geometry_edits.dart
```diff
@@ -574,9 +585,36 @@ class _SnapGeometryEditsState extends State<SnapGeometryEdits>
...
+                  ),
+                  onChanged: (allEnabled) {
+                    if (allEnabled != null) {
+                      allSourceSettings
+                          .map(
+                            (snapSourceSettings) => setState(
+                              () => snapSourceSettings.isEnabled = allEnabled,
+                            ),
+                          )
+                          .toList();
```
PaulAllanSturm, line 611: I don't like using `toList()` for its side-effect (to consume the iterable) while having no use for the returned List. But `forEach` causes a dumb warning and `for (final x in y)` can't be chained. So I dunno, I guess `.toList()` is the best option and is pretty idiomatic in that all Flutter developers end up doing it this way. Unless we want to be really picky and use a `for` loop in this scenario?

PaulAllanSturm (approved):

lib/samples/snap_geometry_edits/snap_geometry_edits.dart
jenmerritt: It is a bit sort of not right isn't it. The forEach thing is really annoying. 
I'm leaning towards leaving it - I think it still feels cleaner this way. Let's see if this arises again and at a non-release time can potentially discuss more.

PR #171, opened 2024-10-12
Style graphics with symbols

martinoyovo:


lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,347 @@
...
+    _createPoints();
+    _createPolygon();
+    _createPolyline();
+    _createText();
+
+    // Update the extent to encompass all of the symbols.
+    _setExtent();
+
+    // Perform some long-running setup task.
+    await Future.delayed(const Duration(seconds: 10));
```
kmueller-gis, line 99: This should be removed

kmueller-gis (commented):
When posting a PR, you should assign reviewers.
- jenmerritt (always include)
- PaulAllenSturm
- kmueller-gis
- changanxian
- harish-kunchala

martinoyovo:
Resolved the comments you added @kmueller-gis!

kmueller-gis (approved):

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,335 @@
...
+
+class StyleGraphicsWithSymbols extends StatefulWidget {
+  const StyleGraphicsWithSymbols({super.key});
+
+  @override
+  State<StyleGraphicsWithSymbols> createState() =>
+      _StyleGraphicsWithSymbolsState();
+}
+
+class _StyleGraphicsWithSymbolsState extends State<StyleGraphicsWithSymbols> {
```
PaulAllanSturm, line 27: This class declaration needs to have `with SampleStateSupport`, and up with the "import" statements it needs
```dart
import '../../utils/sample_state_support.dart';
```

lib/samples/style_graphics_with_symbols/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Allows you to customize a graphic by assigning a unique symbol. For example, you may wish to display individual graphics for different landmarks across a region, and to style each one with a unique symbol.  
+
+## How to use the sample
+
+Pan and zoom around the map. Observe the graphics on the map.
+
+## How it works
+
+1. Create a `GraphicsOverlay` and add it to the `MapView`.
```
PaulAllanSturm, line 17: We rename `MapView` to `ArcGISMapView`. The READMEs have to be updated whenever you see one of our renamed classes.

lib/samples/style_graphics_with_symbols/README.md
```diff
@@ -0,0 +1,37 @@
...
+2. Create a `Symbol` such as `SimpleMarkerSymbol`, `SimpleLineSymbol` or `SimpleFillSymbol`.
```
PaulAllanSturm, line 18: We rename `Symbol` to `ArcGISSymbol` (here and on the next line).

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,335 @@
...
+                child: Center(child: CircularProgressIndicator()),
+              ),
+            ),
+          ),
+        ],
+      ),
+    );
+  }
+
+  void onMapViewReady() async {
```
PaulAllanSturm, line 60: This method doesn't use `await`, so it doesn't need to be declared `async`. Also, `_createPolyline`, `_createPolygon`, `_createText`, and `_setExtent` don't need `async`.

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,335 @@
...
+    // Set the map to the mapview controller.
+    _mapViewController.arcGISMap = map;
+
+    // Add graphics overlay to the mapview controller.
+    _mapViewController.graphicsOverlays.add(_graphicsOverlay);
+
+    // Call functions to create the graphics.
+    _createPoints();
+    _createPolygon();
+    _createPolyline();
```
PaulAllanSturm, line 83: These two lines should be swapped so that they are in the same order as their declarations below (Polyline before Polygon).

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
martinoyovo: Done @PaulAllanSturm, let me know if there's anything else.

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,336 @@
...
+    // Update the extent to encompass all of the symbols.
+    _setExtent();
+
+    // Set the ready state variable to true to enable the sample UI.
+    setState(() => _ready = true);
+  }
+
+  void _createPoints() {
+    // Create a red circle simple marker symbol.
+    SimpleMarkerSymbol redCircleSymbol = SimpleMarkerSymbol(
```
PaulAllanSturm, line 95: One of our style guidelines is to rely on type inference as much as possible. So in this case, we would say:
```dart
final redCircleSymbol = SimpleMarkerSymbol(
```
There are a few other places below where we would use `final` rather than specifying the class name, where it uses `SimpleLineSymbol`, 'SimpleFillSymbol`, `TextSymbol`, and `ArcGISPoint`. (And I think one place we would say `var graphic = Graphic(` because it is reused, so we can't say `final`.)

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,336 @@
...
+        style: SimpleMarkerSymbolStyle.circle, color: Colors.red, size: 10,);
```
PaulAllanSturm, line 96: There's a comma near the end of the line here. Normally that would get reformatted by `dart format`. Normally we let VSCode run the formatter when it saves a file. I'm not sure if you need to configure VSCode to run the formatter, or if maybe it just got missed here for some other reason. But you could run `dart format` from the command line also.

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,336 @@
...
+    final polyline = polylineBuilder.toGeometry();
+
+    // Create the graphic with polyline and symbol.
+    final graphic = Graphic(geometry: polyline, symbol: lineSymbol);
+
+    // Add graphic to the graphics overlay.
+    _graphicsOverlay.graphics.add(graphic);
+  }
+
+  void _createPolygon() async {
```
PaulAllanSturm, line 203: I think there should be a comment here, something like `// Create a green dash line symbol.`

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,336 @@
...
+      symbol: craigleithTextSymbol,
+    );
+
+    // Add graphics to the graphics overlay.
+    _graphicsOverlay.graphics.add(
+      bassRockGraphic,
+    );
+    _graphicsOverlay.graphics.add(
+      craigleithGraphic,
+    );
```
PaulAllanSturm, line 305: You can simplify this by using `addAll`
```dart
_graphicsOverlay.graphics.addAll([bassRockGraphic, craigleithGraphic]);
```

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,336 @@
...
+    // Get all of the graphics contained in the graphics overlay.
+    List<Graphic> myGraphicCollection = _graphicsOverlay.graphics;
+
+    // Create a new envelope builder using the same spatial reference as the graphics.
+    EnvelopeBuilder myEnvelopeBuilder = EnvelopeBuilder(
+      spatialReference: SpatialReference.wgs84,
+    );
+
+    // Loop through each graphic in the graphic collection.
+    for (final Graphic graphic in myGraphicCollection) {
```
PaulAllanSturm, line 318: This could just be `for (final graphic in _graphicsOverlay.graphics) {`

It doesn't need to specify `Graphic` and I don't think the `myGraphicCollection` variable is doing anything useful.

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,336 @@
...
+
+    // Build a polyline.
+    final polylineBuilder =
+        PolylineBuilder(spatialReference: SpatialReference.wgs84);
+    polylineBuilder.addPoint(
+      ArcGISPoint(
+        x: -2.715,
+        y: 56.061,
+      ),
+    );
```
PaulAllanSturm, line 156: There's a more compact way to do this. Instead of `addPoint`, use `addPointXY`
```dart
polylineBuilder.addPointXY(x: -2.715, y:56.061);
```
That'll save a lot of bulk here and in `_createPolygon` too.

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
martinoyovo: Fixed. This is part of the original code we get when we generate the sample. Think we can also update the generator so we can remove the `async` once for all from the samples that will be generated.

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
martinoyovo: I've formatted it with the `dart format` command. Most of the time I use Android Studio to code and VS Code to debug. I'll make sure to configure the formatter in VS Code, that'll help as well.

martinoyovo:
@PaulAllanSturm comments resolved !

lib/samples/style_graphics_with_symbols/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## How to use the sample
+
+Pan and zoom around the map. Observe the graphics on the map.
+
+## How it works
+
+1. Create a `GraphicsOverlay` and add it to the `ArcGISMapView`.
+2. Create a `ArcGISSymbol` such as `SimpleMarkerSymbol`, `SimpleLineSymbol` or `SimpleFillSymbol`.
+3. Create a `Graphic`, specifying a `Geometry` and a `Symbol`.
```
PaulAllanSturm, line 19: This `Symbol` should be `ArcGISSymbol`

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,322 @@
...
+      geometry: polyline,
+      symbol: lineSymbol,
+    );
+
+    // Add graphic to the graphics overlay.
+    _graphicsOverlay.graphics.add(graphic);
+  }
+
+  void _createPolygon() {
+    // Create a green dash line symbol
```
PaulAllanSturm, line 197: There should be a period at the end of the comment

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,322 @@
...
+    // Add graphics to the graphics overlay.
+    _graphicsOverlay.graphics.addAll([
+      bassRockGraphic,
+      craigleithGraphic,
+    ]);
+  }
+
+  void _setExtent() async {
+    // Get all of the graphics contained in the graphics overlay.
+    List<Graphic> myGraphicCollection = _graphicsOverlay.graphics;
```
PaulAllanSturm, line 296: This could just be `final myGraphicCollection = _graphicsOverlay.graphics;`

Or you don't even really need `myGraphicCollection`, you could just use `_graphicsOverlay.graphics` directly in the for loop below, like `for (final graphic in _graphicsOverlay.graphics) {`

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,322 @@
...
+
+    // Create a new envelope builder using the same spatial reference as the graphics.
+    EnvelopeBuilder myEnvelopeBuilder = EnvelopeBuilder(
```
PaulAllanSturm, line 299: Our usual style would be `final myEnvelopeBuilder = EnvelopeBuilder(`

lib/samples/style_graphics_with_symbols/style_graphics_with_symbols.dart
```diff
@@ -0,0 +1,322 @@
...
+      outline: outlineSymbol,
+    );
+
+    // Create a new point collection for polygon.
+    final polygonBuilder =
+        PolygonBuilder(spatialReference: SpatialReference.wgs84);
+    polygonBuilder.addPointXY(
+      x: -2.6425,
+      y: 56.0784,
+    );
```
PaulAllanSturm, line 217: This can all fit on one line by removing the `,` at the end and then letting it get reformatted.
```dart
polygonBuilder.addPointXY(x: -2.6425, y: 56.0784); 
```
That's how we've done these bulk data points in other samples

martinoyovo:
@PaulAllanSturm please check

PaulAllanSturm (approved):

PR #172, opened 2024-10-14
Show WebTiledLayerFromURL Sample Created

harish-kunchala:
Tiled map services are a set of pre-generated images (e.g. "tiles") arranged in folders for each row, column, and zoom level. As you navigate the map, map tiles are requested for the current extent. ArcGISTiledLayer and WMTSLayer are types of tiled map services used for specific data types. WebTiledLayer is useful for displaying other data sources that contain tiles arranged in a row/column/level directory structure, such as OpenStreetMap.


lib/samples/show_web_tiled_layer_from_url/show_web_tiled_layer_from_url.dart
```diff
@@ -0,0 +1,95 @@
...
+
+class _ShowWebTiledLayerFromUrlState extends State<ShowWebTiledLayerFromUrl>
+    with SampleStateSupport {
+  // Create a controller for the map view.
+  final _mapViewController = ArcGISMapView.createController();
+
+  // A flag for when the map view is ready and controls can be used.
+  var _ready = false;
+
+  // Templated URL to the tile service
```
PaulAllanSturm, line 36: Comments should follow the standard of capitalizing the first word and ending with a period. (Here and other comments below.)

lib/samples/show_web_tiled_layer_from_url/show_web_tiled_layer_from_url.dart
```diff
@@ -0,0 +1,95 @@
...
+  final String _templateUri =
```
PaulAllanSturm, line 37: Here (and for `_attribution`) we wouldn't specify `String` because the type gets inferred properly. So it'd just be `final _templateUri =`.

I'm also curious if we can use `const` instead of `final`?

lib/samples/show_web_tiled_layer_from_url/show_web_tiled_layer_from_url.dart
```diff
@@ -0,0 +1,95 @@
...
+              ),
+            ),
+          ],
+        ),
+      ),
+    );
+  }
+
+  void onMapViewReady() async {
+
```
PaulAllanSturm, line 78: we don't need an empty line here

lib/samples/show_web_tiled_layer_from_url/show_web_tiled_layer_from_url.dart
```diff
@@ -0,0 +1,95 @@
...
+    // Create the WebTiledLayer from the URL
+    final webTiledLayer = WebTiledLayer(template: _templateUri)
+    // Set the attribution
+        ..setAttribution(_attribution);
+
+    // Create a Basemap from the WebTiledLayer
+    final basemap = Basemap.withBaseLayer(webTiledLayer);
+
+    // Create a ArcGISMap to hold the BaseMap
```
PaulAllanSturm, line 87: we wouldn't capitalize `BaseMap` like that -- I think we just refer to it as `basemap`

lib/samples/show_web_tiled_layer_from_url/show_web_tiled_layer_from_url.dart
```diff
@@ -0,0 +1,95 @@
...
+                ),
+              ),
+            ),
+          ],
+        ),
+      ),
+    );
+  }
+
+  void onMapViewReady() async {
```
PaulAllanSturm, line 77: we'd only add `async` if the method had any `await` work to do, so it's unnecessary here

lib/samples/show_web_tiled_layer_from_url/show_web_tiled_layer_from_url.dart
PaulAllanSturm: Actually there's no need for these to class members -- they're only used in one place, so they should just be local variables inside `onMapViewReady`.

lib/samples/show_web_tiled_layer_from_url/show_web_tiled_layer_from_url.dart
```diff
@@ -0,0 +1,95 @@
...
+    // Create the WebTiledLayer from the URL
+    final webTiledLayer = WebTiledLayer(template: _templateUri)
+    // Set the attribution
+        ..setAttribution(_attribution);
+
+    // Create a Basemap from the WebTiledLayer
+    final basemap = Basemap.withBaseLayer(webTiledLayer);
+
+    // Create a ArcGISMap to hold the BaseMap
+    final arcGISMap = ArcGISMap.withBasemap(basemap);
```
PaulAllanSturm, line 88: Oh yeah, by convention we would name this variable `map`.

lib/samples/show_web_tiled_layer_from_url/show_web_tiled_layer_from_url.dart
harish-kunchala: I've moved them inside the `onMapViewReady()` and set them as `const` also removed the changed their names from `_templateUrl` and `_attribution` to `templateUrl` and `attribution` as they are local variables now.

lib/samples/show_web_tiled_layer_from_url/show_web_tiled_layer_from_url.dart
harish-kunchala: Made the corrections in comments and verified that all of them follow the rules.

lib/samples/show_web_tiled_layer_from_url/show_web_tiled_layer_from_url.dart
harish-kunchala: Removed and added comment for `onMapViewReady()`

lib/samples/show_web_tiled_layer_from_url/show_web_tiled_layer_from_url.dart
harish-kunchala: Noted. Changed it to `basemap`.

lib/samples/show_web_tiled_layer_from_url/show_web_tiled_layer_from_url.dart
harish-kunchala: Changed `arcGISMap` to `map`.

harish-kunchala:
Made updates according to comments from @PaulAllanSturm.

PR #175, opened 2024-10-15
Implement Change Viewpoint sample

martinoyovo:


.vscode/settings.json
```diff
@@ -0,0 +1,2 @@
+{
+}
```
changanxian, line 2: remove the changes for this file

assets/generated_samples_list.json
```diff
@@ -472,6 +472,38 @@
...
+      "rotate",
+      "scale",
+      "view",
+      "zoom"
+    ],
+    "offline_data": [],
+    "redirect_from": [],
+    "relevant_apis": [
+      "Geometry",
+      "Map",
```
changanxian, line 496: it should be `ArcGISMap`, the file may need to regenerate after modifying the README.

lib/samples/change_viewpoint/change_viewpoint.dart
```diff
@@ -0,0 +1,191 @@
...
+  final _mapViewController = ArcGISMapView.createController();
+
+  // Define the Redlands polygon.
+  late PolygonBuilder _redlandsEnvelope;
+
+  // Define the Edinburgh polygon.
+  late PolygonBuilder _edinburghEnvelope;
+
+  // String array to store titles for the viewpoints specified above.
+  final _viewpointTitles = [
```
changanxian, line 39: Could you use `enum` to define the testing viewpoint types?

martinoyovo: Yes, but wouldn't it look complex from a user point of view ? Since we want to keep the sample simple, I'm wondering if this wouldn't make complex ?

PaulAllanSturm: The samples are not consistent on this -- manage_features.dart uses String for the damageType dropdown.

PaulAllanSturm: Actually manage_features.dart uses both approaches -- damageType is very simple and uses String, and featureManagementOptions uses a custom enum with a `getLabel()` method.

PaulAllanSturm: As a suggestion, we could use a String here for now, and open a Discussion on whether we should always use a custom enum for dropdown menus.

martinoyovo:
Comments resolved @PaulAllanSturm @changanxian.

changanxian (approved):
👍

PaulAllanSturm (approved):

PR #177, opened 2024-10-17
Control Annotation Sublayer Visibility

martinoyovo:


lib/samples/control_annotation_sublayer_visibility/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Use case
+
+Annotation, which differs from labels by having a fixed place and size, is typically only relevant
+at particular scales. Annotation sublayers allow for finer control of annotation by allowing
+properties (like visibility in the map and legend) to be set and others to be read (like name) on
+subtypes of an annotation layer.
+
+An annotation dataset which marks valves as "Opened" or "Closed", might be set to display the "
+Closed" valves over a broader range of scales than the "Opened" valves, if the "Closed" data is
```
jenmerritt, line 15:
```suggestion
An annotation dataset which marks valves as "Opened" or "Closed", might be set to display the "Closed" valves over a broader range of scales than the "Opened" valves, if the "Closed" data is
```

lib/samples/control_annotation_sublayer_visibility/README.md
```diff
@@ -0,0 +1,59 @@
...
+considered more relevant by the map's author. Regardless, the user can be given a manual option to
+set visibility of annotation sublayers on and off, if required.
+
+## How to use the sample
+
+Start the sample and take note of the visibility of the annotation. Zoom in and out to see the
+annotation turn on and off based on scale ranges set on the data.
+
+Use the checkboxes to manually set "Open" and "Closed" annotation sublayers visibility to on or off.
```
jenmerritt, line 24:
```suggestion
Use the switches to manually set "Open" and "Closed" annotation sublayers visibility to on or off.
```

lib/samples/control_annotation_sublayer_visibility/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## How it works
+
+1. Create a new `ArcGISMapView` with an `ArcGISMapViewController`.
```
jenmerritt, line 28:
```suggestion
```
Since this step isn't in the design readme I don't think we need to include it (our comments in the code is sufficient). Note you'll need to amend the numbering accordingly.

lib/samples/control_annotation_sublayer_visibility/README.md
```diff
@@ -0,0 +1,59 @@
...
+2. Load a `MobileMapPackage` that contains `AnnotationSublayer`.
+3. Get the sublayers from the map package's layers by calling `sublayer.subLayerContents`.
+4. You can toggle the visibility of each sublayer manually using `sublayer.isVisible` property.
+5. To determine if a sublayer is visible at the current scale of the `ArcGISMapView`,
+   use `sublayer.isVisibleAtScale(ArcGISMapViewController.scale)`, by passing in the map's current
+   scale.
```
jenmerritt, line 34:
```suggestion
3. Get the sublayers from the map package's layers by calling `Sublayer.subLayerContents`.
4. You can toggle the visibility of each sublayer manually using `Sublayer.isVisible` property.
5. To determine if a sublayer is visible at the current scale of the `ArcGISMapView`,
   use `Sublayer.isVisibleAtScale(ArcGISMapViewController.scale)`, by passing in the map's current
   scale.
```
This is a very minor thing and not incorrect as such, but for consistency we (Flutter) tend to use the Class name here rather than variable. There are some differences in these parts of the readme's between platforms.

lib/samples/control_annotation_sublayer_visibility/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+* AnnotationLayer
+* AnnotationSublayer
+* LayerContent
+
+## Offline data
+
+This sample
+uses [Gas Device Anno Mobile Map Package](https://arcgisruntime.maps.arcgis.com/home/item.html?id=b87307dcfb26411eb2e92e1627cb615b).
+It is downloaded from ArcGIS Online automatically before the sample runs.
```
jenmerritt, line 46:
```suggestion
This sample uses [Gas Device Anno Mobile Map Package](https://arcgisruntime.maps.arcgis.com/home/item.html?id=b87307dcfb26411eb2e92e1627cb615b).
It is downloaded from ArcGIS Online as part of the sample.
```

lib/samples/control_annotation_sublayer_visibility/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## About the data
+
+The scale ranges were set by the map's author using ArcGIS Pro:
+
+* The "Open" annotation sublayer has its maximum scale set to 1:500 and its minimum scale set to 1:
+  2000.
+* The "Closed" annotation sublayer has no minimum or maximum scales set, so will be drawn at all
+  scales.
```
jenmerritt, line 55:
```suggestion
* The "Open" annotation sublayer has its maximum scale set to 1:500 and its minimum scale set to 1:2000.
* The "Closed" annotation sublayer has no minimum or maximum scales set, so will be drawn at all scales.
```

lib/samples/control_annotation_sublayer_visibility/control_annotation_sublayer_visibility.png
jenmerritt, line 1: We need to recapture this without the "Beta - For Developer Use Only" watermark. There should be no license watermark visible in the screengrabs. Once we're out of Beta the watermark will become "Licensed for Developer Use Only" and we'll also need to remove that for our sample screengrabs.

jenmerritt (commented):
Mostly focused on the readme rather than in-depth code review.

lib/samples/control_annotation_sublayer_visibility/control_annotation_sublayer_visibility.png
PaulAllanSturm: My usual process for removing the Beta watermark is to comment out this line in arcgis_environment.dart:
```dart
  // Set the Beta watermark
  runtimecore.RT_ArcGISRuntimeEnvironment_setBetaWatermark(true, ffi.nullptr);
```
You have to be running the Sample Viewer against your development "flutter" repo.

To get rid of the "Licensed for Developer Use Only" message, I hard-code the license from env.json into sample_runner.dart.

Then I revert those changes without checking them in to git.

jenmerritt: I often just crop it out for a sample like this where there is no UI at the top, but I'll improve the internal doc to explain how to remove fully (rather than link to a conversation as before which is less helpful).

lib/samples/control_annotation_sublayer_visibility/control_annotation_sublayer_visibility.dart
```diff
@@ -0,0 +1,231 @@
...
+        _mapViewController.onViewpointChanged.listen((_) {
+          // Check if the sublayer is visible at the current map scale.
+          if (_openSublayer!.isVisibleAtScale(_mapViewController.scale)) {
+            _openLabelColor = Colors.purple;
+          } else {
+            _openLabelColor = null;
+          }
+          // Set the current map scale text.
+          _currentScaleLabel =
+              'Current map scale: 1:${_mapViewController.scale.toInt()}';
```
PaulAllanSturm, line 221: `round()` is probably slightly better than `toInt()` here

PaulAllanSturm: Probably we should additionally set `_currentScaleLabel` right away as part of `onMapViewReady`, rather than waiting for the first `viewpointChanged` event to happen. I think the information is available immediately, so it's kind of weird to leave it blank at the beginning.

lib/samples/control_annotation_sublayer_visibility/control_annotation_sublayer_visibility.dart
```diff
@@ -0,0 +1,231 @@
...
+
+          setState(() {});
```
PaulAllanSturm, line 223: This works, but `setState(() {});` is frowned upon because it doesn't tell us what has changed. Ideally there'd be a `setState` wrapped around each line of code that impacts the widget.

lib/samples/control_annotation_sublayer_visibility/control_annotation_sublayer_visibility.dart
```diff
@@ -0,0 +1,231 @@
...
+            ),
+          Text(
+            _currentScaleLabel ?? '',
+          ),
+        ],
+      ),
+    );
+  }
+
+  void onMapViewReady() async {
```
PaulAllanSturm, line 172: So we're still a little fuzzy on how we want to do error checking for data that has been specially prepared for a sample. If you look at what we did in `manage_features.dart`, that is probably the best approach we have so far. It has a big try/catch in its `onMapViewReady`. In the body of the `try`, it doesn't check for null and it doesn't check for exceptions thrown by `firstWhere`. Any such problems are caught by the two clauses at the bottom -- one for ArcGISException, and one catch-all.

```dart
  void onMapViewReady() async {
    try {
// ... setting up all the data ...
    } on ArcGISException catch (e) {
      showMessageDialog('${e.message}.');
    } on Exception {
      showMessageDialog(
        'There was an error loading the data required for the sample.',
      );
    }
```
I think we should take the same approach here. That allows the code to read a little better, without having the isNotEmpty and null checks slowing things down.

lib/samples/control_annotation_sublayer_visibility/README.md
martinoyovo: Thank you for pointing it out !

lib/samples/control_annotation_sublayer_visibility/control_annotation_sublayer_visibility.png
martinoyovo: @jenmerritt I saw your PR set the license key of the `ArcGISEnvironment`, in order to remove the watermark. I've used that method and it removes the watermark as expected. Thanks for the suggestion !

martinoyovo: I'll update the screenshot as well.

lib/samples/control_annotation_sublayer_visibility/control_annotation_sublayer_visibility.dart
```diff
@@ -0,0 +1,249 @@
...
+      await downloadSampleData(['b87307dcfb26411eb2e92e1627cb615b']);
+      final appDir = await getApplicationDocumentsDirectory();
+
+      // Load the mobile map package.
+      final mmpkFile = File('${appDir.absolute.path}/GasDeviceAnno.mmpk');
+      // Mobile map package that contains annotation layers.
+      final mmpk = MobileMapPackage.withFileUri(mmpkFile.uri);
+      await mmpk.load();
+
+      if (mmpk.maps.isNotEmpty) {
```
PaulAllanSturm, line 177: We can get rid of the check for isNotEmpty now. If it throws, we have it covered by the try/catch block.

lib/samples/control_annotation_sublayer_visibility/control_annotation_sublayer_visibility.dart
```diff
@@ -0,0 +1,249 @@
...
+        // Get the first map in the mobile map package and set to the map view.
+        _mapViewController.arcGISMap = mmpk.maps.first;
+      }
+
+      if (_mapViewController.arcGISMap != null) {
```
PaulAllanSturm, line 182: Likewise we no longer need to check arcGISMap for null -- we know it has been set.

lib/samples/control_annotation_sublayer_visibility/control_annotation_sublayer_visibility.dart
```diff
@@ -0,0 +1,249 @@
...
+              '${_openSublayer.name} (1:${_openSublayer.maxScale.toInt()} - 1:${_openSublayer.minScale.toInt()})';
+        });
+
+        // Add event handler for changing the text color to indicate whether the "open" sublayer is visible at the current scale.
+        _mapViewController.onViewpointChanged.listen((_) {
+          // Check if the sublayer is visible at the current map scale.
+          if (_openSublayer.isVisibleAtScale(_mapViewController.scale)) {
+            setState(() {
+              _openLabelColor = Colors.purple;
+            });
```
PaulAllanSturm, line 213: We don't have a hard rule about this, but this can fit on one line so we should prefer that
```dart
setState(() => _openLabelColor = Colors.purple);
```
and likewise below ...
```dart
setState(() => _openLabelColor = null);
```

PaulAllanSturm (approved):

PR #183, opened 2024-10-25
Set up for use with pub.dev

PaulAllanSturm:
Configure pubspec.yaml to use pub.dev and update the readme


jenmerritt (approved):
I think that's good! 👍

PR #186, opened 2024-10-31
Simple "About" page

PaulAllanSturm:
A simple "About" page with the version number. This required moving the "MaterialApp" widget up to the "runApp" call.

Screenshot 2024-10-31 at 3 33 27 PM

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -31,7 +31,15 @@ void main() {
   } else {
     ArcGISEnvironment.apiKey = apiKey;
   }
-  runApp(const SampleViewerApp());
+
+  runApp(
+    MaterialApp(
+      theme: ThemeData(
+        colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
+      ),
+      home: const SampleViewerApp(),
+    ),
+  );
 }
 
 class SampleViewerApp extends StatefulWidget {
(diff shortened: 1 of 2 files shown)


jenmerritt (commented):
Only thought is UX wise if it could be an 'X' to close/dismiss the widget rather than a back button. The arrow implies to me that we're navigating through that popup.

PaulAllanSturm:
> Only thought is UX wise if it could be an 'X' to close/dismiss the widget rather than a back button. The arrow implies to me that we're navigating through that popup.

done

jenmerritt (approved):
Will need a dependency request too

PaulAllanSturm:
> Will need a dependency request too

package_info_plus is already on our Approved list for the core package. Do we need a separate request for the Sample Viewer?

jenmerritt:
Ohh no no, that's all good, I hadn't actually checked 👍

martinoyovo:
Can we update the ui design and use an AlertDialog instead ?

martinoyovo (commented):
Can we update the ui design and use an AlertDialog instead ?

PaulAllanSturm:
> Can we update the ui design and use an AlertDialog instead ?

We intend to add more information here. It won't all fit in an AlertDialog, especially in landscape mode. We're essentially aiming for something like the About screen in the Swift SDK's Sample Viewer:

Screenshot 2024-11-04 at 2 18 05 PM

martinoyovo:
Clear now @PaulAllanSturm !

PR #188, opened 2024-11-05
Apply style to WMS Layer

martinoyovo:


lib/samples/apply_style_to_wms_layer/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+![Image of apply style to WMS layer](apply_style_to_wms_layer.png)
+
+## Use case
+
+Layers hosted on WMS may have different pre-set styles available to apply to them. Swapping between these styles can help during visual examination of the data. For example, increasing the contrast of satellite images can help in identifying urban and agricultural areas within forested areas.
+
+## How to use the sample
+
+Once the layer loads, the toggle button will be enabled. Click it to toggle between the first and second styles of the WMS layer.
```
jenmerritt, line 13:
```suggestion
Once the layer loads, toggle between the first and second styles of the WMS layer using the dropdown menu.
```
Or words to that effect that better describe our UI.

lib/samples/apply_style_to_wms_layer/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+## How it works
+
+1. Create a `WmsLayer` specifying the URL of the service and the layer names you want `WmsLayer.withUriAndLayerNames(uri: Uri.parse(url), layerNames: names)`.
+2. When the layer is done loading, get one of its sublayer `wmsLayer.sublayers.first as WmsSublayer`.
+3. After obtaining one of its sub layers, get its list of style strings using `firstSublayer.sublayerInfo.styles`.
+4. Set one of the styles using `firstSublayer.currentStyle = styles[0]`.
```
jenmerritt, line 20: `wmsLayer.sublayers.first as WmsSublayer`. This cast had me down a rabbit hole of API ref! It looks like the sample design is based on another implementation - there seems to be some minor difference between the sdks here with whether sublayers themselves actually come back from that sublayers property. It looks like our API is correct (per the API design) but I think this is what you want to do in this scenario for our API:

When the layer is done loading, get its list of style strings using `wmsLayer.layerInfos.first.styles`.

Then ideally we won't display a cast in the readme steps so maybe the next step could end up being something like:
Set one of the styles to the first sublayer using `WmsSublayer.currentStyle = style.first`.

lib/samples/apply_style_to_wms_layer/apply_style_to_wms_layer.dart
```diff
@@ -0,0 +1,164 @@
+// Copyright 2024 Esri
```
jenmerritt, line 1:
```suggestion
//
// Copyright 2024 Esri
```
I think the copyright is missing something. It should get generated correctly, if it was incorrect in the generation we may need to revisit. Or if it was a copy-paste thing then that's fine, just correct here.

lib/samples/apply_style_to_wms_layer/apply_style_to_wms_layer.dart
```diff
@@ -0,0 +1,164 @@
...
+    );
+  }
+
+  void changeStyle(String style) async {
+    // Set the selected style.
+    setState(() => _selectedStyle = style);
+
+    // Get the available styles from the first sublayer.
+    final firstSublayer = _wmsLayer.sublayers.first as WmsSublayer;
+    final styles = firstSublayer.sublayerInfo.styles;
```
jenmerritt, line 120: I think we can follow my comment in the readme to instead get the styles from the `layerInfos` property on the `WmsLayer` etc.

lib/samples/apply_style_to_wms_layer/apply_style_to_wms_layer.dart
```diff
@@ -0,0 +1,164 @@
...
+        setState(() => firstSublayer.currentStyle = styles[1]);
+      default:
+        throw StateError('Unknown style');
+    }
+  }
+
+  void onMapViewReady() async {
+    // Create a map with spatial reference appropriate for the service.
+    final map = ArcGISMap(spatialReference: SpatialReference(wkid: 26915));
+    map.minScale = 7000000.0;
```
jenmerritt, line 137:
```suggestion
    final map = ArcGISMap(spatialReference: SpatialReference(wkid: 26915))
    ..minScale = 7000000.0;
```
Maybe an opportunity for cascade notation? Unless you have any thoughts on its usage generally?

lib/samples/apply_style_to_wms_layer/apply_style_to_wms_layer.dart
```diff
@@ -0,0 +1,164 @@
...
+    // Create a new WMS layer displaying the specified layers from the service.
+    // The default styles are chosen by default.
+    _wmsLayer = WmsLayer.withUriAndLayerNames(
+      uri: Uri.parse(
+        'https://imageserver.gisdata.mn.gov/cgi-bin/mncomp?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities',
+      ),
+      layerNames: ['mncomp'],
+    );
```
jenmerritt, line 145:
```suggestion
   const wmsLayerUri = 'https://imageserver.gisdata.mn.gov/cgi-bin/mncomp?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities';
    _wmsLayer = WmsLayer.withUriAndLayerNames(
      uri: Uri.parse(wmsLayerUri),
      layerNames: ['mncomp'],
    );
```
This is a nit-pick and not incorrect, but follows a pattern we've been using in other samples.

lib/samples/apply_style_to_wms_layer/README.md
martinoyovo: I think it's missing `.sublayerInfo` before `.styles` in :  `wmsLayer.layerInfos.first.styles`.

lib/samples/apply_style_to_wms_layer/apply_style_to_wms_layer.dart
martinoyovo: In this case, it's incorrectly generated. Most of the samples follow the previous format. I've fixed it, particularly for this sample but we may need to revisit the generation as you said.

lib/samples/apply_style_to_wms_layer/apply_style_to_wms_layer.dart
martinoyovo: Great suggestion, thank you!

lib/samples/apply_style_to_wms_layer/README.md
```diff
@@ -0,0 +1,33 @@
...
+Layers hosted on WMS may have different pre-set styles available to apply to them. Swapping between these styles can help during visual examination of the data. For example, increasing the contrast of satellite images can help in identifying urban and agricultural areas within forested areas.
+
+## How to use the sample
+
+Once the layer loads, toggle between the first and second styles of the WMS layer using the dropdown menu.
+
+## How it works
+
+1. Create a `WmsLayer` specifying the URL of the service and the layer names you want `WmsLayer.withUriAndLayerNames(uri: Uri.parse(url), layerNames: names)`.
+2. When the layer is done loading, get its list of style strings using `wmsLayer.layerInfos.styles`.
```
jenmerritt, line 18:
```suggestion
2. When the layer is done loading, get its list of style strings using `WmsLayer.layerInfos.styles`.
```

lib/samples/apply_style_to_wms_layer/apply_style_to_wms_layer.dart
```diff
@@ -0,0 +1,165 @@
...
+        onChanged: (style) {
+          if (style != null) {
+            changeStyle(style);
+          }
+        },
+      ),
+    );
+  }
+
+  void changeStyle(String style) async {
```
PaulAllanSturm, line 115: this method doesn't need to be `async`

martinoyovo: I used `await` to load the `_wmsLayer`:
```
// Wait for the layer to load.
    await _wmsLayer.load();
```

PaulAllanSturm: `await _wmsLayer.load()` is down in `onMapViewReady`. Here in `changeStyle`, there are no `await` calls.

martinoyovo: oh fact, my bad!

PaulAllanSturm (approved):

PR #195, opened 2024-11-19
cs/Adds a SnackBar notification

changanxian:
- Adds a SnackBar for notification when the download is finished

### DIFF
--- a/lib/samples/download_vector_tiles_to_local_cache/download_vector_tiles_to_local_cache.dart
+++ b/lib/samples/download_vector_tiles_to_local_cache/download_vector_tiles_to_local_cache.dart
@@ -207,6 +207,27 @@ class _DownloadVectorTilesToLocalCacheState
       },
     );
 
+    _exportVectorTilesJob?.onStatusChanged.listen(
+      (status) {
+        if (status == JobStatus.succeeded) {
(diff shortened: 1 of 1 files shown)


lib/samples/download_vector_tiles_to_local_cache/download_vector_tiles_to_local_cache.dart
```diff
@@ -207,6 +207,27 @@ class _DownloadVectorTilesToLocalCacheState
...
+                content: const Text('Downloaded vector tiles successfully'),
+                duration: const Duration(seconds: 5),
+                backgroundColor: Theme.of(context).colorScheme.secondary,
+              ),
+            );
+          }
+          // Dismiss the progress indicator.
+          setState(() {
+            _isJobStarted = false;
+          });
```
PaulAllanSturm, line 226: This could be the shorter version:
```dart
setState(() => _isJobStarted = false);
```
Also, if the job fails, the progress indicator will never be dismissed. Maybe this needs to go back where it was or we should check for `JobStatus.failed` here.

martinoyovo (approved):

PaulAllanSturm (approved):

jenmerritt (approved):

PR #197, opened 2024-11-21
Cs/2566 update creating mobile geodatabase sample

changanxian:
- avoid the text width overflow on certain device.


changanxian:
close it wrong branch

PR #208, opened 2024-12-05
Use build_runner to generate certain files

PaulAllanSturm:
Use the `build_runner` package to generate `assets/generated_samples_list.json` and `lib/models/samples_widget_list.dart`.

Instead of having these files checked in to the repository, they are instead created by running `dart run build_runner build`.


lib/builder.dart
```diff
@@ -0,0 +1,132 @@
+import 'dart:convert';
```
changanxian, line 1: missing copyright

pubspec.yaml
```diff
@@ -21,9 +21,13 @@ dependencies:
   simple_html_css: ^5.0.0
   webview_flutter: ^4.8.0
   package_info_plus: ^8.1.0
+  build: ^2.4.1
```
changanxian, line 24: this can be 2.4.12

README.md
```diff
@@ -26,6 +26,12 @@ Note on Windows: this step requires permission to create symlinks. Either run th
 dart run arcgis_maps install
 ```
 
+Use `build_runner` to generate support code.
+
+```
+dart run build_runner build
+```
+
```
jenmerritt, line 34: I guess I didn't realise this would be another step for our users which I'm on the fence about. Not totally opposed but would prefer not since the benefits of doing it is really to aid our team. 
Could we maybe have a single command for the samples that does the install and build steps?

PaulAllanSturm: We could present it as:

```sh
dart run arcgis_maps install && dart run build_runner build
```

And/or we could drop in a script that does those two steps (as well as `flutter pub upgrade`). Call it `initialize.sh` maybe.

pubspec.yaml
PaulAllanSturm: I only see 2.4.1 as the latest: https://pub.dev/packages/build

changanxian (approved):

README.md
PaulAllanSturm: ok, I added a script and edited the README

PaulAllanSturm:
The build_runner code doesn't work on Windows (due to the usual `/` vs `\` problems). I need to fix that and then get fresh reviews.

PaulAllanSturm:
In order to support building on Windows, I made an `initialize.bat` to go along with `initialize.sh`.

changanxian (approved):

README.md
jenmerritt: Nice. My only thought now is if we should call that shell/bat from a dart script so our users call `dart run initialize` or something. Can you raise this in parking lot - I'm interested in thoughts from others in the team on this experience. It's quite a key entry point so I just want to give it the thought and discussion.

PaulAllanSturm:
The initialize step is now `dart tool/initialize.dart` on both platforms.

README.md
PaulAllanSturm: `dart tool/initialize.dart`

jenmerritt: I think this works well. 
My only query now before approving is whether for a publicly consumed script it should be in `bin` rather than an internal tool. So it would be like: 
`dart run arcgis_maps_sdk_flutter_samples initialize`
I guess the lines are a little blurred. Interested in your thoughts.

jenmerritt: `dart run arcgis_maps_sdk_flutter_samples initialize` is maybe a little more descriptive of what it's doing too as side effect.

PaulAllanSturm: I think this is definitely an internal tool. If you're building the Sample Viewer, you are stepping inside the project to build it (and the project is otherwise useless until this is run). The `bin` folder would be for functionality that is being provided for some purpose outside of the project.

jenmerritt (approved):

PR #215, opened 2024-12-16
Cs/refactor msg dialog

changanxian:
- Refactoring the common AlertDialog into a common usage.

### DIFF
--- a/lib/common/dialogs.dart
+++ b/lib/common/dialogs.dart
@@ -14,21 +14,31 @@
 // limitations under the License.
 //
 
-import 'package:arcgis_maps_sdk_flutter_samples/common/common.dart';
 import 'package:flutter/material.dart';
 
-/// A mixin that overrides `setState` to first check if the widget is mounted.
-/// (Calling `setState` on an unmounted widget causes an exception.)
-mixin SampleStateSupport<T extends StatefulWidget> on State<T> {
-  @override
-  void setState(VoidCallback fn) {
-    if (mounted) super.setState(fn);
-  }
-
-  /// Shows an alert dialog with the given [message].
-  void showMessageDialog(String message, {String title = 'Info', bool showOK = false}) {
-    if (mounted) {
-      showAlertDialog(context,  message, title: title, showOK: showOK);
-    }
-  }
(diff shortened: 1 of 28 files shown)


lib/samples/show_service_area/show_service_area.dart
```diff
@@ -260,20 +253,7 @@ class _ShowServiceAreaState extends State<ShowServiceArea>
...
-          ),
-          actions: [
-            TextButton(
-              onPressed: () => Navigator.pop(context, 'OK'),
-              child: const Text('OK'),
-            ),
-          ],
-        ),
-      );
+      showMessageDialog('At least 1 facility is required to perform a service area calculation.', showOK: true,);
```
PaulAllanSturm, line 256: run `dart format .` to get the desired formatting (here and probably elsewhere too)

lib/common/dialogs.dart
```diff
@@ -0,0 +1,51 @@
...
+//
+
+import 'package:flutter/material.dart';
+
+/// 
+/// Shows an alert dialog with a message.
+/// - [context]: The context in which the dialog is shown.
+/// - [message]: The message to display in the dialog.
+/// - [title]: (Optional) The title of the dialog.
+/// - [showOK]: (Optional) A boolean value to determine if the dialog should show an OK button.
```
PaulAllanSturm, line 24: The parameter name is "displayOkButton", not "showOK". Probably it should be "showOK" for consistency with sample_state_support.dart.

lib/common/dialogs.dart
```diff
@@ -0,0 +1,51 @@
...
+
+Future<void> showAlertDialog(
+  BuildContext context,
+  String message, {
+  String title = 'Alert',
+  bool displayOkButton = false,
+}) {
+  // if the context State object is in the widget tree.
```
PaulAllanSturm, line 32: I don't understand this comment.

lib/common/dialogs.dart
changanxian: updated

PaulAllanSturm (approved):

PR #216, opened 2024-12-18
Cs/Refactoring theme data and adding Category main UI 

changanxian:
In this changes,

- A category UI is added to categorize the samples based on the common design:

Now the main page is as 

When the Category card is clicked, e.g,  `Analysis`

The filtered samples are displayed as before, the search function still work.

- A centralized theme data is created to hold all theme related data, `sampleViewerTheme`, 
most of theme related data are refactored in the class, but maybe some cases are missed.


lib/main.dart
```diff
@@ -33,60 +34,24 @@ void main() {
...
+  final sampleViewerPage = SampleViewerPage();
 
   @override
   Widget build(BuildContext context) {
-    const title = 'ArcGIS Maps SDK for Flutter Samples';
-
     return Scaffold(
       appBar: AppBar(
-        title: const Text(title),
+        title: const Text('Sample Categories'),
```
jenmerritt, line 54:
```suggestion
        title: const Text('Categories'),
```

lib/widgets/category_card.dart
```diff
@@ -0,0 +1,115 @@
...
+              children: [
+                CategoryIconBackground(category: category),
+                Text(
+                  category.title,
+                  style: const TextStyle(
+                    color: Colors.white,
+                    fontSize: 16.0,
+                    fontWeight: FontWeight.bold,
+                  ),
+                  //textAlign: TextAlign.center,
```
jenmerritt, line 55:
```suggestion
```

jenmerritt (commented):
A few initial thoughts. I need to spend more time thinking about this as there is quite a bit to consider. 

Categories:
- I think the title when you select a category should be "category_name" not "ArcGIS Maps SDK for Flutter samples"
- We should have an "All" option so that a user can view and search all the samples. As right now it's a bit restrictive having to guess what category a sample is in before searching.

Theme:
- We could consider doing the category view and theme in separate PRs, or if this PR is just applying a theme and not changing any styling then that might be manageable.
- The theme colors ultimately need to be updated to match the official Runtime purple - if not as part of this then we'll want a follow up issue.

For all reviewers benefits, there isn't a physical "common design" for the sample viewer as such, though certainly some patterns exist across all SDKs. So we do have some freedom to think about how we best want the UI to work for Flutter, within some restrictions of what we have from the individual sample common designs.

changanxian:
>We could consider doing the category view and theme in separate PRs, or if this PR is just applying a theme and not changing any styling then that might be manageable.

Yes, I agree. I should create the separate PRs for the theme data and category UI. The initial goal is to use theme data and not change our current styling, and updatable in one central place.

We should discuss category UI change separately.

jenmerritt:
> The initial goal is to use theme data and not change our current styling, and updatable in one central place.

👍  that sounds like a good plan. Whatever is easiest PR wise.

PR #217, opened 2024-12-18
Cs/Refactoring the theme data

changanxian:


lib/samples/add_feature_layers/add_feature_layers.dart
```diff
@@ -102,20 +102,18 @@ class _AddFeatureLayersState extends State<AddFeatureLayers>
...
-                ),
+                style: Theme.of(context).textTheme.labelMedium,
               ),
               // Set the selected feature layer source.
               value: _selectedFeatureLayerSource,
-              icon: const Icon(
+              icon: Icon(
                 Icons.arrow_drop_down,
-                color: Colors.deepPurple,
+                color: Theme.of(context).colorScheme.primary,
```
PaulAllanSturm, line 113: Can we use `ThemeData.iconTheme` to specify the color as part of theme? Then we don't need to set the `color:` property of every `Icon`. We could use simply `icon: Icon(Icons.arrow_drop_down)` and the color will be automatically set from the `ThemeData`.

lib/samples/create_and_edit_geometries/create_and_edit_geometries.dart
```diff
@@ -399,7 +393,7 @@ class _CreateAndEditGeometriesState extends State<CreateAndEditGeometries>
                   Tooltip(
                     message: 'Undo',
                     child: ElevatedButton(
-                      style: _buttonStyle,
+                      style: Theme.of(context).elevatedButtonTheme.style,
```
PaulAllanSturm, line 396: Is this redundant? Won't an ElevatedButton automatically pick up the ThemeData's elevatedButtonTheme? I think we can remove the `style:` property of all the `ElevatedButton`s and they'll automatically inherit the proper theme.

lib/samples/set_basemap/set_basemap.dart
```diff
@@ -124,7 +124,7 @@ class _SetBasemapState extends State<SetBasemap> with SampleStateSupport {
             child: FloatingActionButton(
               onPressed: () => _scaffoldStateKey.currentState!.openEndDrawer(),
               shape: const RoundedRectangleBorder(),
-              backgroundColor: Colors.deepPurple,
+              backgroundColor: Theme.of(context).primaryColor,
```
PaulAllanSturm, line 127: I think we can use `ThemeData.floatingActionButtonTheme` to specify both `shape` and `backgroundColor` for all `FloatingActionButton`s. Then we don't have to specify these properties for every `FloatingActionButton`.

lib/samples/add_feature_layers/add_feature_layers.dart
```diff
@@ -102,20 +102,18 @@ class _AddFeatureLayersState extends State<AddFeatureLayers>
...
               ),
               elevation: 16,
-              style: const TextStyle(color: Colors.deepPurple),
+              style: Theme.of(context).textTheme.labelMedium,
```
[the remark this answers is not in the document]
changanxian: There is no such way to call TextTheme like: `TextTheme.of(context).labelMedium`

lib/common/theme_data.dart
```diff
@@ -0,0 +1,64 @@
...
+        fontWeight: FontWeight.bold,
+        color: Colors.white,
+      );
+
+  TextStyle get customErrorStyle => const TextStyle(color: Colors.red);
+  TextStyle get customWhiteStyle => const TextStyle(color: Colors.white);
+}
+
+extension CustomColorScheme on ColorScheme {
+  Color get circleColor => const Color(0xFFFF0000);
```
PaulAllanSturm, line 63: I don't see where `circleColor`, `customLabelStyle`, and `customErrorStyle` are used. Should they be deleted?

jenmerritt (approved):
When this is merged, make sure to make Kathleen / the release team aware so they have it in mind when doing their testing.

PaulAllanSturm (approved):

PR #219, opened 2024-12-30
Adopt very_good_analysis 7.0.0 linter rules

PaulAllanSturm:
The very_good_analysis project provides a list of recommended lint rules: https://github.com/VeryGoodOpenSource/very_good_analysis/blob/main/lib/analysis_options.7.0.0.yaml

We want to defer to their expert opinion and adopt these rules. Turning on these rules necessitated numerous changes to our code.

Some of the rules we intentionally turn off, for reasons documented in the linter_rules.yaml file.


jenmerritt (approved):
Ok with me if this makes things more defined. One thing that occurred to me is it might be worth checking the sample skeleton to make sure it meets the new requirements. It may just be the `Future<void> onMapViewReady()`.

PaulAllanSturm:
> One thing that occurred to me is it might be worth checking the sample skeleton to make sure it meets the new requirements. It may just be the `Future<void> onMapViewReady()`.

Yes that's in progress.

PR #223, opened 2025-01-04
Show network status on sample page

martinoyovo:


jenmerritt (commented):
I like the find, in terms of the package that identifies network connection, but I'm not sure if the implementation is quite right.

The original issue talks about displaying a message when downloading in Generate offline map samples as the progress bar just shows 0 so it seems like nothing is happening. For that part, I'm not sure if the message stands out enough. Perhaps we'd be better showing a dialog when that button is tapped?



This PR is adding a notification in all samples, which is potentially a useful thing but the UI placement of the message overlaps with UI from other samples e.g. "Create buffers around points" and "Identify features in WMS layer". We could make the background opaque but then that would interfere with samples that work offline e.g. if you turned the network connection off while using "download preplanned map area" to play with displaying the offline maps, this message would hide part of the active UI:



We could go down the route Swift have where we don't display the sample at all and it's a full height, centered widget. The downside with this is that users can't use offline workflow samples at all if they were exploring that...

If we're implementing an app wide thing we might need to create a separate issue and give it some more thought.

PaulAllanSturm:
There are definitely two distinct use cases:

1. A general app-wide notification of network status
2. Error handling when a particular network call fails

A solution for # 1 isn't enough by itself, because a network call might fail even if the network is up. I.e., the network might be up, but our particular web service might be down.

To solve # 2, we need to handle the error condition when the call fails. I'm not sure what the error condition is in this particular case. We might be catching an exception and then ignoring it. Instead of ignoring it, we would need to present an alert dialog with the error message.

jenmerritt:
After reviewing the scope of this task and discussing with both @PaulAllanSturm and @martinoyovo , we'll focus on addressing the behavior in the Generate Offline Map sample when the `_generateOfflineMapJob!.run();` method fails - either due to no network connectivity (or it could be due to the service being down, for example).

Another issue captures us more widely addressing network errors in the samples - for the various potential causes - as well as providing an indication of network status to users somewhere within the sample viewer.

@martinoyovo - up to you if you prefer a fresh PR. This one will be useful for future reference and consideration of the connectivity_plus package even if it's closed.

PR #228, opened 2025-01-10
Cs/Add category UI

changanxian:
Adds a new Category UI following the discussion, and Flow layout to organize the CategrorCard Widgets.
The new UI looks like this now:

in Portait mode:

in Landscape mode:


lib/models/category.dart
```diff
@@ -0,0 +1,117 @@
...
+  layers('Layers'),
+  maps('Maps'),
+  routingAndLogistics('Routing and Logistics'),
+  // scenes('Scenes'),
+  searchAndQuery('Search and Query'),
+  // utilityNetworks('Utility Networks'),
+  visualization('Visualization');
+  // favorites('Favorites');
+  const SampleCategory(this.title);
+  final String title;
```
PaulAllanSturm, line 47: I think some code can be simplified here. We could add "icon" and "backgroundImage" to `enum SampleCategory`, and then get rid of `class Category` entirely.

```dart
enum SampleCategory {
  all('All', Icons.apps, 'assets/category_images/all_background.webp');
  // ... all the other values ...
  const SampleCategory(this.title, this.icon, this.backgroundImage);
  final String title;
  final IconData icon;
  final String backgroundImage;
}
```

Then, we could get rid of the `sampleCategories` list and just use `SampleCategory.values` in its place.

PaulAllanSturm: Also that way we can stop using `import ... as arcgis_category;`, because there won't be a collision with the name of the `Category` class.

lib/main.dart
```diff
@@ -50,40 +49,14 @@ class SampleViewerApp extends StatefulWidget {
...
-    _searchFocusNode.dispose();
-    super.dispose();
-  }
+  static const double cardSpacing = 6;
 
   @override
   Widget build(BuildContext context) {
-    const title = 'ArcGIS Maps SDK for Flutter Samples';
-
+    const applicationTitle = 'ArcGIS Maps SDK Flutter Samples';
```
jenmerritt, line 56:
```suggestion
    const applicationTitle = 'ArcGIS Maps SDK for Flutter Samples';
```

lib/widgets/category_card.dart
```diff
@@ -0,0 +1,112 @@
...
+                mainAxisAlignment: MainAxisAlignment.center,
+                children: [
+                  CategoryIconBackground(category: category),
+                  Text(
+                    category.title,
+                    style: const TextStyle(
+                      color: Colors.white,
+                      fontSize: 16,
+                      fontWeight: FontWeight.bold,
+                    ),
```
jenmerritt, line 54: This jumped out as inline styling. Is there any themeData we can use here?

lib/widgets/sample_viewer_page.dart
```diff
@@ -0,0 +1,166 @@
...
+    _textEditingController.dispose();
+    _searchFocusNode.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: const Text('ArcGIS Maps SDK Flutter Samples'),
```
jenmerritt, line 66:
```suggestion
        title: const Text('ArcGIS Maps SDK for Flutter Samples'),
```

jenmerritt (commented):
The category page itself and the search from there is very slick.

My main issue right now is that once you enter a category (or All) you can't search, which I think is a detrimental user experience. 

I think we either need the search icon on the internal pages too - though this should limit the search to the current category.
Or I think my preference would be that once you select a category the search bar is just at the top like the previous UI. And again, this should limit the search to the current category.

lib/widgets/sample_viewer_page.dart
```diff
@@ -0,0 +1,166 @@
...
+  State<SampleViewerPage> createState() => _SampleViewerPageState();
+}
+
+class _SampleViewerPageState extends State<SampleViewerPage> {
+  final _allSamples = <Sample>[];
+  final _searchFocusNode = FocusNode();
+  final _textEditingController = TextEditingController();
+  var _filteredSamples = <Sample>[];
+  bool _ready = false;
+  bool _searchHasFocus = false;
```
PaulAllanSturm, line 40: Our general style is to declare these `var` instead of `bool`

PaulAllanSturm:
> Or I think my preference would be that once you select a category the search bar is just at the top like the previous UI. And again, this should limit the search to the current category.

I agree with this.

changanxian:
Make the improvement on search:

-> Search is always available on the top
-> if a category is selected, it only searches under the category
-> all is as before

jenmerritt:
Works well! Just a couple more thoughts to refine:
- When we load the "icon" search, can this initially display all the samples (rather than none?) i.e. if no search term is provided I think it should default to all rather than blank (more useful)
- Titles: 
   - On the homepage, can we make it "Sample Categories" rather than just "Categories". Kotlin have this and actually I think that is a little more descriptive so think we should follow suit.
   - When you select a category, can the title on that screen be the selected category? Rather than "ArcGIS Maps SDK for Flutter Samples" - I think it helps with knowing where you've navigated to, otherwise it's hard to remember what category you selected.

Thanks!

lib/widgets/sample_viewer_page.dart
```diff
@@ -132,22 +132,41 @@ class _SampleViewerPageState extends State<SampleViewerPage> {
...
-    final List<Sample> results;
+    var results = <Sample>[];
+    // restore the initial list of samples if the search text is empty
     if (searchText.isEmpty) {
-      results = _allSamples;
+      if (widget.category == null) {
+        results = [];
+      } else if (widget.category == SampleCategory.all) {
+        results = _allSamples;
+      } else if (widget.category != null) {
```
PaulAllanSturm, line 142: This can just be `} else {` (with no `if` condition) because the case where `widget.category == null` was already handled in the first branch of the `if` statement. So we know it can't be `null` here.

lib/widgets/sample_viewer_page.dart
```diff
@@ -132,22 +132,41 @@ class _SampleViewerPageState extends State<SampleViewerPage> {
...
+            final lowerSearchText = searchText.toLowerCase();
+            return sample.title.toLowerCase().contains(lowerSearchText) ||
+                sample.category.toLowerCase().contains(lowerSearchText) ||
+                sample.keywords.any(
+                  (keyword) => keyword.toLowerCase().contains(lowerSearchText),
+                );
+          },
+        ).toList();
+      // if the category is not null, the only samples within the category are searched
+      } else if (widget.category != null) {
```
PaulAllanSturm, line 158: Likewise the `if (widget.category != null)` is redundant here, because the first branch of the `if` statement is taking care of the case where `widget.category == null`.

lib/widgets/category_card.dart
```diff
@@ -0,0 +1,109 @@
...
+//
+
+import 'package:arcgis_maps_sdk_flutter_samples/common/common.dart';
+import 'package:arcgis_maps_sdk_flutter_samples/models/category.dart';
+import 'package:flutter/material.dart';
+
+/// A card widget that displays a SampleCategory.
+class CategoryCard extends StatelessWidget {
+   const CategoryCard({
+    required this.category, required this.onClick, super.key,
```
PaulAllanSturm, line 24: I think `dart format` needs to be run on this file. I think it would put these arguments on separate lines.

lib/widgets/sample_viewer_page.dart
```diff
@@ -0,0 +1,185 @@
...
+    } else {
+      _searchFocusNode.requestFocus();
+    }
+  }
+
+  Future<void> loadSamples() async {
+    final jsonString =
+        await rootBundle.loadString('assets/generated_samples_list.json');
+    final sampleData = jsonDecode(jsonString);
+    for (final s in sampleData.entries) {
```
PaulAllanSturm, line 122: One of the new lint rules is triggered by this line. To fix the lint warning, we have to call `jsonDecode` like this:

```dart
    final sampleData = jsonDecode(jsonString) as Map<String, dynamic>;
```

changanxian:
Nice feedback, updated.

PaulAllanSturm (approved):

jenmerritt:
Looks great! Except this comment:

> When we load the "icon" search, can this initially display all the samples (rather than none?) i.e. if no search term is provided I think it should default to all rather than blank (more useful)

So IMO the search from the category page should behave like "All". What do you think? Maybe we just need to initially set the filtered samples list to all samples or something?

changanxian:
In my opinion, the Icon search is different from the All search:

- In the All search, a predefined list of samples is shown, and the search is conducted based on the title and keywords.
- In the Icon search, there is no predefined list of samples (similar to a Google search bar). Instead, the results are generated dynamically by searching the title, keywords, and descriptions (which is not done in the All search).
Currently, our sample viewer does not have the functionality to display the full description page of each sample (this might be a separate issue). Also, in the Icon search, I would like to highlight the keywords in the selected samples that match the search terms.

I can create a separate issue to enhance the <Icon> search and how to display the full descriptions of each sample, we may discuss there. What do you think?

PaulAllanSturm:
I think it makes sense that there are two different search workflows.
1. When you click on a category (or "All"), you immediately see the list and can optionally choose to filter the list. If you're just browsing, you come here, and you can ignore the text field for filtering if you want. ("Browse and filter".)
2. When you click on the search icon, you're doing something different -- you're coming here with the intention of searching, so you have a search field and you only see matches to your search query. ("Keyword search.")

jenmerritt:
I think a separate issue is good. It works fine for now and we can come back to when we want to re-prioritize the sample viewer.

jenmerritt (approved):

PR #232, opened 2025-01-21
Cs/update flutter to 3.27.2

changanxian:
- Update the Flutter SDK to use the default `3.27.2` and Dart '3.6.1'
- Fixes the issues reported from flutter analyze

### DIFF
--- a/lib/samples/control_annotation_sublayer_visibility/control_annotation_sublayer_visibility.dart
+++ b/lib/samples/control_annotation_sublayer_visibility/control_annotation_sublayer_visibility.dart
@@ -174,7 +174,8 @@ class _ControlAnnotationSublayerVisibilityState
       _mapViewController.onViewpointChanged.listen((_) {
         // Check if the sublayer is visible at the current map scale.
         if (_openSublayer.isVisibleAtScale(_mapViewController.scale)) {
-          setState(() => _openLabelColor = Theme.of(context).colorScheme.primary);
+          setState(
+              () => _openLabelColor = Theme.of(context).colorScheme.primary);
         } else {
           setState(() => _openLabelColor = null);
         }
--- a/lib/samples/edit_with_branch_versioning/edit_with_branch_versioning.dart
+++ b/lib/samples/edit_with_branch_versioning/edit_with_branch_versioning.dart
@@ -91,7 +91,7 @@ class _EditWithBranchVersioningState extends State<EditWithBranchVersioning> wit
               child: IgnorePointer(
                 child: Container(
                   padding: const EdgeInsets.all(10),
-                  color: Colors.white.withOpacity(0.7),
+                  color: Colors.white.withValues(alpha: 0.7),
                   child: Row(
                     mainAxisAlignment: MainAxisAlignment.center,
                     children: [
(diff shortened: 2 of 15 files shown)


lib/samples/control_annotation_sublayer_visibility/control_annotation_sublayer_visibility.dart
```diff
@@ -174,7 +174,8 @@ class _ControlAnnotationSublayerVisibilityState
       _mapViewController.onViewpointChanged.listen((_) {
         // Check if the sublayer is visible at the current map scale.
         if (_openSublayer.isVisibleAtScale(_mapViewController.scale)) {
-          setState(() => _openLabelColor = Theme.of(context).colorScheme.primary);
+          setState(
+              () => _openLabelColor = Theme.of(context).colorScheme.primary,);
```
PaulAllanSturm, line 178: This needs `dart doc` to format it properly.

lib/samples/edit_with_branch_versioning/edit_with_branch_versioning.dart
```diff
@@ -358,43 +358,47 @@ class _EditWithBranchVersioningState extends State<EditWithBranchVersioning> wit
...
-                                            title: const Text('Error'),
-                                            content: Text(
-                                              'Error: $cleanedMessage',
-                                            ),
-                                            actions: [
-                                              TextButton(
-                                                onPressed: () =>
-                                                    Navigator.of(context).pop(),
-                                                child: const Text('OK'),
+                                      setState(() async {
```
PaulAllanSturm, line 364: Why was `setState()` added here? It seems odd that we would wrap a `showDialog` in a `setState()`.

lib/samples/edit_with_branch_versioning/edit_with_branch_versioning.dart
```diff
@@ -453,23 +457,27 @@ class _EditWithBranchVersioningState extends State<EditWithBranchVersioning> wit
...
-                          actions: [
-                            TextButton(
-                              onPressed: () => Navigator.of(context).pop(),
-                              child: const Text('OK'),
-                            ),
-                          ],
-                        );
-                      },
-                    );
+                    setState(() async {
```
PaulAllanSturm, line 463: Also here -- I don't understand the `setState()`

lib/samples/edit_with_branch_versioning/edit_with_branch_versioning.dart
changanxian: The reason I put the block into `setState` is to avoid the lint warning issue:
```
Don't use 'BuildContext's across async gaps, guarded by an unrelated 'mounted' check.
Guard a 'State.context' use with a 'mounted' check on the State, and other BuildContext use with a 'mounted' check on the BuildContext.dart[use_build_context_synchronously](https://dart.dev/diagnostics/use_build_context_synchronously)
```

Even I put a `if (mounted)` check, the warning still shows up.   what do you think to get rid of the warnings? or I leave the warnings as is for this PR.

changanxian: warnings

PaulAllanSturm: I think I figured it out. Instead of `if (mounted)`, you have to do `if (context.mounted)`. This is because we're using a `StatefulBuilder` widget, which passes in a `context` parameter. So we have to check the `mounted` value on that `context` object, rather than the `mounted` value that comes from the `State` class.

PaulAllanSturm (approved):

martinoyovo (approved):

PR #240, opened 2025-01-24
Remove comment from issue template

jenmerritt:
Need to remove the comment as formatting impacting the issue template.


jenmerritt:
Overriding merge since it isn't within the sample code and just config.

PR #248, opened 2025-02-06
Identify raster cell sample

martinoyovo:


lib/samples/identify_raster_cell/README.md
```diff
@@ -0,0 +1,35 @@
...
+2. On tap:
+  * Call `ArcGISMapViewController.identifyLayer(...)` passing in the screen point, tolerance, and maximum number of results per layer.
+  * Await the result of the identify and then get the `GeoElement` from the layer result.
+  * Create a `SimpleRenderer` at the calculated map point and populate the content with text from the `GeoElement.attributes`.
+
+## Relevant API
+
+* ArcGISMapViewController.identifyLayer(...)
+* GeoElement
+* GeoElement.attributes(...)
```
PaulAllanSturm, line 25: There shouldn't be "(...)" here because `attributes` isn't a method call.

I'm not sure if we should use "(...)" with `identifyLayer(...)`. I think we should remove the "(...)" there, too. I haven't seen any other sample README.md use "(...)".

lib/samples/identify_raster_cell/README.metadata.json
```diff
@@ -0,0 +1,40 @@
...
+        "cell value",
+        "continuous",
+        "discrete",
+        "identify",
+        "pixel",
+        "pixel value",
+        "raster",
+        "ArcGISMapViewController.identifyLayer(...)",
+        "GeoElement",
+        "GeoElement.attributes(...)",
```
PaulAllanSturm, line 21: If we remove "(...)" from README.md, we'll have to make the same changes here and in the "relevant_apis" section below.

lib/samples/identify_raster_cell/identify_raster_cell.dart
```diff
@@ -0,0 +1,148 @@
...
+            // Display a progress indicator and prevent interaction until state is ready.
+            LoadingIndicator(visible: !_ready),
+          ],
+        ),
+      ),
+    );
+  }
+
+  Future<void> onMapViewReady() async {
+    // Define a new map with Wgs84 Spatial Reference.
```
PaulAllanSturm, line 66: I don't think this comment is quite accurate. I think we would normally just say,
```
  // Create a map with the oceans basemap style and an initial viewpoint.
```

lib/samples/identify_raster_cell/identify_raster_cell.dart
```diff
@@ -0,0 +1,148 @@
...
+    _rasterLayer = RasterLayer.withRaster(raster);
+    // Load the Raster Layer.
+    await _rasterLayer.load();
+    // Add the Raster Layer to the map.
+    map.operationalLayers.add(_rasterLayer);
+    // Set the viewpoint.
+    if (_rasterLayer.fullExtent != null) {
+      await _mapViewController.setViewpointGeometry(_rasterLayer.fullExtent!);
+    }
+    _textGraphicsOverlay.graphics.add(_textGraphic);
```
PaulAllanSturm, line 99: There should be a comment above this line to describe what these two lines are doing.

lib/samples/identify_raster_cell/identify_raster_cell.dart
```diff
@@ -0,0 +1,148 @@
...
+  // Graphics overlay to display the text graphic.
+  final _textGraphicsOverlay = GraphicsOverlay();
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      body: SafeArea(
+        top: false,
+        left: false,
+        right: false,
```
harish-kunchala, line 49: You can remove Safe Area

lib/samples/identify_raster_cell/identify_raster_cell.dart
PaulAllanSturm: Actually, there's probably no reason to set the `initialViewpoint`, because the viewpoint will be changed immediately with the `_rasterLayer.fullExtent`. So we could just remove the `initialViewpoint` part and the above comment could just be "// Create a map with the oceans basemap style."

lib/samples/identify_raster_cell/identify_raster_cell.dart
```diff
@@ -0,0 +1,148 @@
...
+      await _mapViewController.setViewpointGeometry(_rasterLayer.fullExtent!);
+    }
+    _textGraphicsOverlay.graphics.add(_textGraphic);
+    _mapViewController.graphicsOverlays.add(_textGraphicsOverlay);
+    // Set the ready state variable to true to enable the sample UI.
+    setState(() => _ready = true);
+  }
+
+  Future<void> onTap(Offset position) async {
+    // Get the result for where the user hovered on the raster layer.
```
PaulAllanSturm, line 106: I don't think "hovered" is correct here -- shouldn't we just say "tapped"?

lib/samples/identify_raster_cell/identify_raster_cell.dart
```diff
@@ -0,0 +1,148 @@
...
+    final identifyResult = await _mapViewController.identifyLayer(
+      _rasterLayer,
+      screenPoint: position,
+      tolerance: 1,
+    );
+    // Create a StringBuffer to display information to the user.
+    final stringBuffer = StringBuffer();
```
PaulAllanSturm, line 113: This variable declaration should be moved inside the "if" statement, because it is not used anywhere else.

lib/samples/identify_raster_cell/identify_raster_cell.dart
```diff
@@ -0,0 +1,148 @@
...
+    // Create a Raster from the local tif file.
+    final raster = Raster.withFileUri(
+      Uri.file(
+        '${appDir.absolute.path}/SA_EVI_8Day_03May20/SA_EVI_8Day_03May20.tif',
+      ),
+    );
+    // Create a Raster Layer.
+    _rasterLayer = RasterLayer.withRaster(raster);
+    // Load the Raster Layer.
+    await _rasterLayer.load();
```
harish-kunchala, line 92: You can put this in another function like `Future<RasterLayer> loadRasterLayer()`, to keep our `onMapViewReady` concise.

lib/samples/identify_raster_cell/identify_raster_cell.dart
```diff
@@ -0,0 +1,148 @@
...
+      // Get the x and y values of the cell.
+      if (cell.geometry != null) {
+        final x = cell.geometry!.extent.xMin;
+        final y = cell.geometry!.extent.yMin;
+        // Add the x & y coordinates where the user clicked raster cell to the string buffer.
+        stringBuffer.writeln();
+        stringBuffer.writeln('X: ${x.toStringAsFixed(4)}');
+        stringBuffer.write('Y: ${y.toStringAsFixed(4)}');
+
+        final textSymbol = TextSymbol(
```
PaulAllanSturm, line 130: @jenmerritt - here's an interesting idea to use a TextSymbol instead of a Callout (until we implement Callout). We haven't used this before, but it makes sense here.

lib/samples/identify_raster_cell/identify_raster_cell.dart
martinoyovo: Correct! Thanks for catching that!

lib/samples/identify_raster_cell/README.md
martinoyovo: Apparently when I run the `tool/readme_scripts/readme_scripts_runner.dart`, it asks me to add the "(...)".

lib/samples/identify_raster_cell/README.metadata.json
martinoyovo: Screenshot 2025-02-06 at 4 03 59 PM

martinoyovo: By removing "(...)" from both the README and the metadata, it works!

lib/samples/identify_raster_cell/README.md
```diff
@@ -0,0 +1,35 @@
+# Identify raster cell
+
+Get the cell value of a local raster at the tapped location and display the result in a callout.
+
+![Image of identify raster cell](identify_raster_cell.png)
+
+## Use case
```
jenmerritt, line 7: This needs populating from the design

lib/samples/identify_raster_cell/README.md
```diff
@@ -0,0 +1,35 @@
+# Identify raster cell
+
+Get the cell value of a local raster at the tapped location and display the result in a callout.
```
jenmerritt, line 3:
```suggestion
Get the cell value of a local raster at the tapped location and display the result in the UI.
```
Callout is a specific API that we will eventually implement but don't support yet.

lib/samples/identify_raster_cell/identify_raster_cell.dart
jenmerritt: Looks good. I don't think we need a rule as to what we suggest instead of Callout but it's another nice option as well as other UI elements 👍

lib/samples/identify_raster_cell/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+![Image of identify raster cell](identify_raster_cell.png)
+
+## Use case
+
+## How to use the sample
+
+Tap an area of the raster to identify it and see the raw raster cell information displayed in a renderer. Tap and hold to see cell information updated dynamically as you drag.
```
jenmerritt, line 11: > Tap and hold to see cell information updated dynamically as you drag.

I don't think this part of the sample has been implemented yet. The Swift sample is a good example - I quite like that they've also enabled the magnifier, but see what you think is the best way for us to implement.

lib/samples/identify_raster_cell/identify_raster_cell.dart
PaulAllanSturm: We only use a SafeArea when there is a row of buttons below the map (or other controls). There aren't any controls for this sample, so we don't need the SafeArea.

lib/samples/identify_raster_cell/README.md
martinoyovo: I can add a magnifier but we don't currently a way to get the position of a magnifier while interacting with the map. We have "onLongPressEnd", but that only returns the offset of when a drag has been released. So I think we won't be able to include this part in the implementation, unless there's another suggestion of doing that.

jenmerritt: Could we use a GestureDetector here? It could be a cool way to have a demo of GestureDetector for users if it's doable. Or is the problem still that we can't get the offset? 🤔  

We could create this as a follow up issue. I'm just trying to think if our users wanted to do something, like drag around and query, how they could do it. We deliberately didn't implement loads of gesture callbacks directly on the map view so a gesture detector could be implemented for whatever the user wanted to do.

jenmerritt: I think the magnifier is only useful if we can query on the fly while dragging around

martinoyovo: Exactly, the magnifier is only useful while dragging around. The idea of using GestureDetector sounds great. I think I can play around with that and see how it goes.

martinoyovo: So after a couple of testings, `GestureDetector` works but doesn't give the expected behavior. When we wrap the `ArcGISMapView` widget, can we get the pointer position but it doesn't follow the magnifer direction at the same time, which makes it a bit complicated to track the position when the magnifier is enabled. Also, I found that `RawGestureDetector` is used in the `ArcGISMapView`, probably to enable the long press and the tap on the map, so I think we can extend that to support other gestures in the future.

I'm not confident about the behavior of when `ArcGISMapView` is wrapped in the `GestureDetector`. I suggest we we keep the implementation without the magnifier enabled and remove that part off the README.

jenmerritt: We have had a lot of discussions about other gestures, I'll share that with you directly. This is a useful use case for us to look at vs other SDKs though. So for now we'll implement just identify on tap in this sample and I'll write this up separately.

lib/samples/identify_raster_cell/README.md
jenmerritt: @martinoyovo this comment still needs to be addressed

lib/samples/identify_raster_cell/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Get the cell value of a local raster at the tapped location and display the result in the UI.
+
+![Image of identify raster cell](identify_raster_cell.png)
+
+## Use case
+
+## How to use the sample
+
+Tap an area of the raster to identify it and see the raw raster cell information displayed in a renderer.
```
jenmerritt, line 11:
```suggestion
Tap an area of the raster to identify it and see the raw raster cell information displayed in the UI.
```
I don't think renderer quite works.

lib/samples/identify_raster_cell/README.md
martinoyovo: No problem, I understand that!

martinoyovo:
Should be fixed now @jenmerritt, thanks for catching that!

PR #252, opened 2025-02-13
Har14220/add raster from file

harish-kunchala:
- Fixed File not loading issue.


jenmerritt (approved):
Ah yes - it will default back to the map's initial viewpoint. An alternative would be to set the map's initial viewpoint. But what you've done here is good and the way we usually do it 👍

kmueller-gis (approved):

PR #253, opened 2025-02-13
Navigate route with rerouting sample

changanxian:
- Implements the navigate route with rerouting sample to test the navigation API


lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,532 @@
...
+                ),
+                Row(
+                  mainAxisAlignment: MainAxisAlignment.spaceEvenly,
+                  children: [
+                    Container(
+                      decoration: BoxDecoration(
+                        shape: BoxShape.circle,
+                        color: Theme.of(context).primaryColorLight,
+                      ),
+                      child: IconButton(
```
jenmerritt, line 113:
```suggestion
                      // Add a button to stop navigation if it has started.
                      child: IconButton(
```

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,532 @@
...
+                        color: Colors.white,
+                        icon: const Icon(Icons.stop),
+                      ),
+                    ),
+                    Container(
+                      decoration: BoxDecoration(
+                        shape: BoxShape.circle,
+                        color: Theme.of(context).primaryColorLight,
+                      ),
+                      child: IconButton(
```
jenmerritt, line 124:
```suggestion
                      // Add a button to stop navigation if it has not already started.
                      child: IconButton(
```

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,532 @@
...
+                        color: Colors.white,
+                        icon: const Icon(Icons.play_arrow),
+                      ),
+                    ),
+                    Container(
+                      decoration: BoxDecoration(
+                        shape: BoxShape.circle,
+                        color: Theme.of(context).primaryColorLight,
+                      ),
+                      child: IconButton(
```
jenmerritt, line 135:
```suggestion
                      // Add a button to toggle navigation mode.
                      child: IconButton(
```

jenmerritt: I find the `needRecenter` state variable a little confusing. Could the variable instead of being a boolean, be a `LocationDisplayAutoPanMode` and then this button only be enabled if it `!= LocationDisplayAutoPanMode.Navigation` ?

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,532 @@
...
+  Future<SimulatedLocationDataSource> getLocationDataSource() async {
+    final routePointJson =
+        await rootBundle.loadString('assets/SanDiegoTourPath.json');
+    final routeLine = Geometry.fromJsonString(routePointJson) as Polyline;
+
+    final simulatedLocationDataSource = SimulatedLocationDataSource()
+      ..setLocationsWithPolyline(
+        routeLine,
+        simulationParameters: SimulationParameters(
+          speed: 40,
```
jenmerritt, line 456: I can see why you've increased the speed, but I think it's a little too quick for the built up area and also that the voice isn't as effective since there are deliberately so many instructions / recalculations right at the start. Maybe go with 20 so it's a little quicker than the default but still feels less rushed.

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,532 @@
...
+
+class NavigateRouteWithRerouting extends StatefulWidget {
+  const NavigateRouteWithRerouting({super.key});
+
+  @override
+  State<NavigateRouteWithRerouting> createState() =>
+      _NavigateRouteWithReroutingState();
+}
+
+class _NavigateRouteWithReroutingState extends State<NavigateRouteWithRerouting>
```
jenmerritt, line 33: General comment - it's quite slow to load - one thing to consider, are there any of the initial `async` operations that aren't sequential and could be loaded alongside one another with a `Future.wait(future1, future2)` as opposed to `await future1; await future2`.

jenmerritt (commented):
Cool to see this coming to life! I played around on iOS simulator and Android device.

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,563 @@
...
+    super.dispose();
+  }
+
+  // Set up the map with a navigation basemap style.
+  Future<void> onMapViewReady() async {
+    // Create a map with a navigation basemap style.
+    final map = ArcGISMap.withBasemapStyle(BasemapStyle.arcGISNavigation);
+    _mapViewController.arcGISMap = map;
+
+    map.onLoadStatusChanged.listen(
```
PaulAllanSturm, line 237: I'm not sure if it's important for the subscription returned by `listen` to be captured by a variable so it can be canceled in the `dispose()`. But that can be avoided if this was written differently:
```dart
map.load().then((_) async {
  setState(() => _initNavigation = true);
  await initRouteTask();
  setState(() => _initNavigation = false);
});
```
I think that's over all simpler.

PaulAllanSturm (commented):
I didn't get through the whole thing yet but here's as far as I got

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,563 @@
...
+  @override
+  void dispose() {
+    _flutterTts.stop();
+    _simulatedLocationDataSource?.stop();
+
+    super.dispose();
+  }
+
+  // Set up the map with a navigation basemap style.
+  Future<void> onMapViewReady() async {
```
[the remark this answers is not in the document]
changanxian: but I got a warning for using async without Future.

assets/SanDiegoTourPath.json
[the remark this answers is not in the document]
changanxian: yes, updated

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
changanxian: `40` is from design, I changed it to 35,  which Koltin sample also use the number.

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
changanxian: I added another indicator to load data and initialize navigation. The UI performance is improved, give a try,

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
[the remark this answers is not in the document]
PaulAllanSturm: The `await map.load().then...` shouldn't have the `await`. When you remove the `await`, you can remove the `async`, and when you remove the `async`, you can change the `Future<void>` to just `void`.

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
PaulAllanSturm: The two downloads can be started earlier and run simultaneously.

In `initState()`, we could do:
```dart
  _geodatabaseFuture = downloadSanDiegoGeodatabase();
  _tourPathFuture = downloadSanDiegoTourPath();
```
Then,  where the Future return values are needed,
```dart
final geodatabasePath = await _geodatabaseFuture;
```
(and the same for the other Future where its value is needed)

PaulAllanSturm: Actually, instead of `_tourPathFuture`, you could have:
```dart
_locationDataSourceFuture = getLocationDataSource();
```

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
jenmerritt: Oh yeah so it is - I did review the design while I was thinking about this but for some reason didn't see it before, interesting. Ok 35 is fine then - hopefully there aren't too many speed cameras 😄

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
changanxian: updated

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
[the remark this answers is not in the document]
changanxian: updated

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,587 @@
...
+  var _reset = false;
+  var _setupDataAndInitNavigation = false;
+  // Variables to show the remaining distance, time, and next direction.
+  var _routeStatus = '';
+  var _remainingDistance = '';
+  var _remainingTime = '';
+  var _nextDirection = '';
+  // Future to download the geodatabase.
+  late Future<String> _geodatabasePathFuture;
+  // Future to download the simulate the location data source.
```
PaulAllanSturm, line 80: "the simulate the location ..." should be "the simulated location ..."

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,587 @@
...
+  }
+
+  // Set up the map with a navigation basemap style.
+  void onMapViewReady() {
+    // Create a map with a navigation basemap style.
+    final map = ArcGISMap.withBasemapStyle(BasemapStyle.arcGISNavigation);
+    _mapViewController.arcGISMap = map;
+
+    
+
```
PaulAllanSturm, line 253: There are three blank lines here. It should just be one blank line.

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,587 @@
...
+    map.onLoadStatusChanged.listen((loadStatus) async{
+      if (loadStatus == LoadStatus.loaded) {
+        setState(() => _setupDataAndInitNavigation = true);
+      await initRouteTask();
+      setState(() => _setupDataAndInitNavigation = false);
```
PaulAllanSturm, line 258: The whitespace needs to be fixed by running `dart format`

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,587 @@
...
+      Stop(_startPoint)..name = 'San Diego Convention Center',
+      Stop(_endPoint)..name = 'RH Fleet Aerospace Museum',
+    ]);
+
+    // Solve the route and store the result.
+    _routeResult = await _routeTask.solveRoute(routeParameters);
+
+    // Initializes and adds graphics to the map view to visually represent the route,
+    // including the remaining route, traveled route, and start/end points.
+    showRouteGraphics();
```
PaulAllanSturm, line 294: Maybe this method should be called "initRouteGraphics()"?

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,587 @@
...
+        setState(() => _resetToNavigationMode = true);
+      } else {
+        setState(() => _resetToNavigationMode = false);
+      }
+    });
+
+    // Update the remaining route graphic and center the map view on the route.
+    await zoomToRoute();
+
+    // Set the route tracker locale
```
PaulAllanSturm, line 346: period at the end of this comment

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,587 @@
...
+    _routeTracker.voiceGuidanceUnitSystem =
+        const Locale.fromSubtags().languageCode == 'en'
+            ? UnitSystem.metric
+            : UnitSystem.imperial;
```
PaulAllanSturm, line 350: Should this logic be the other way around? Most countries would use metric.

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,587 @@
...
+      }
+    });
+
+    setState(() => _ready = true);
+  }
+
+  // Initialize the route task, route parameters, and route result.
+  Future<void> initRouteTask() async {
+    final geodatabasePath = await _geodatabasePathFuture;
+    _simulatedLocationDataSource = await _simulatedLocationDataSourceFuture;
```
PaulAllanSturm, line 268: `_simulatedLocationDataSource` isn't needed until later in the process, down in `initNavigation()`. We should move this to where  `_simulatedLocationDataSource` first gets used. That way, the Future has more time to complete.

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,587 @@
...
+
+  // Set up the map with a navigation basemap style.
+  void onMapViewReady() {
+    // Create a map with a navigation basemap style.
+    final map = ArcGISMap.withBasemapStyle(BasemapStyle.arcGISNavigation);
+    _mapViewController.arcGISMap = map;
+
+    
+
+    map.onLoadStatusChanged.listen((loadStatus) async{
```
kmueller-gis, line 254: This stream subscription is not cancelled.

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,587 @@
...
+    final routeTrackerLocationDataSource = RouteTrackerLocationDataSource(
+      routeTracker: _routeTracker,
+      locationDataSource: _simulatedLocationDataSource,
+    );
+    // Set the location data source.
+    _mapViewController.locationDisplay.dataSource =
+        routeTrackerLocationDataSource;
+    _mapViewController.locationDisplay.autoPanMode =
+        LocationDisplayAutoPanMode.navigation;
+    _mapViewController.locationDisplay.onAutoPanModeChanged.listen((event) {
```
kmueller-gis, line 335: StreamSubscription is not cancelled.

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,587 @@
...
+            ? UnitSystem.metric
+            : UnitSystem.imperial;
+
+    // Listen for voice guidance and tracking status changes.
+    _routeTracker.onNewVoiceGuidance.listen(updateGuidance);
+    _routeTracker.onTrackingStatusChanged.listen(updateProgress);
+    _routeTracker.onRerouteStarted.listen((_) {
+      setState(() => _routeStatus = 'Rerouting');
+    });
+    _routeTracker.onRerouteCompleted.listen((_) {
```
kmueller-gis, line 358: Route Tracker subscriptions are not getting cancelled.

kmueller-gis:
As far as the StreamSubscriptions, we have been inconsistent in cancelling them in our samples. In other samples, there have been some subscriptions that have introduced leaks, but I don't see any memory leaks resulting from these subscriptions.

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,587 @@
...
+  Future<void> initNavigation() async {
+    // Create the route tracker with rerouting enabled.
+    _routeTracker = RouteTracker.create(
+      routeResult: _routeResult,
+      routeIndex: 0,
+      skipCoincidentStops: true,
+    )!;
+
+    // Enable rerouting on the route tracker.
+    if (_routeTask.getRouteTaskInfo().supportsRerouting) {
```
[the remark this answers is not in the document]
changanxian: I remove the `return`,  if the service is not supporting rerouting, it still continue with the original route. thank

kmueller-gis (approved):

lib/samples/navigate_route_with_rerouting/navigate_route_with_rerouting.dart
```diff
@@ -0,0 +1,600 @@
...
+  var _nextDirection = '';
+  // Future to download the geodatabase.
+  late Future<String> _geodatabasePathFuture;
+  // Future to download the simulated location data source.
+  late Future<SimulatedLocationDataSource> _simulatedLocationDataSourceFuture;
+  // Stream subscriptions
+  late StreamSubscription<VoiceGuidance> _voiceGuidanceSubscription;
+  late StreamSubscription<TrackingStatus> _trackingStatusSubscription;
+  late StreamSubscription<void> _rerouteStartedSubscription;
+  late StreamSubscription<void> _rerouteCompletedSubscription;
```
PaulAllanSturm, line 87: Declaring these `late` is problematic because "dispose()" might get called before they are set, which will cause an exception and crash. For example, if the network downloads are slow and the user hits "Back" before all the "init" methods run. So these StreamSubscriptions should be declared without `late` and as nullable.

PaulAllanSturm (approved):

PR #265, opened 2025-02-25
Add internationalization support to sample viewer.

martinoyovo:


pubspec.yaml
```diff
@@ -29,6 +31,9 @@ dev_dependencies:
   build_runner: ^2.4.15
   flutter_lints: ^5.0.0
 
+dependency_overrides:
```
martinoyovo, line 34: Without `dependency_overrides` there's a version conflict between the `intl` package versions required by `flutter_localizations` and `arcgis_maps`.

pubspec.yaml
```diff
@@ -16,6 +16,8 @@ dependencies:
   flutter:
     sdk: flutter
   flutter_archive: ^6.0.3
+  flutter_localizations:
+    sdk: flutter
```
changanxian, line 20: I don't understand why the package need to be a part of `flutter` library instead of a specific version?

martinoyovo: The flutter_localization package is [part of Flutter's API](https://api.flutter.dev/flutter/flutter_localizations/). Also, they recommend setting it up that way: https://docs.flutter.dev/ui/accessibility-and-internationalization/internationalization#setting-up.

changanxian (approved):

PaulAllanSturm (approved):

PR #266, opened 2025-03-03
cs/replace sizebox with spacing

changanxian:
- Replace SizedBox used in the Row or Column with the spacing property.

### DIFF
--- a/lib/samples/show_legend/show_legend.dart
+++ b/lib/samples/show_legend/show_legend.dart
@@ -133,13 +133,13 @@ class _ShowLegendState extends State<ShowLegend> with SampleStateSupport {
           DropdownMenuItem(
             value: legend,
             child: Row(
+              spacing: 8,
               children: [
                 // Add the legend image to the dropdown list if the image exists.
                 if (arcGISImage != null)
                   Image.memory(arcGISImage.getEncodedBuffer())
                 else
                   Container(),
-                const SizedBox(width: 8),
                 // Add the legend name to the dropdown list.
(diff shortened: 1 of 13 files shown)


lib/samples/show_legend/show_legend.dart
```diff
@@ -133,13 +133,13 @@ class _ShowLegendState extends State<ShowLegend> with SampleStateSupport {
...
             value: legend,
             child: Row(
+              spacing: 8,
               children: [
                 // Add the legend image to the dropdown list if the image exists.
                 if (arcGISImage != null)
                   Image.memory(arcGISImage.getEncodedBuffer())
                 else
                   Container(),
-                const SizedBox(width: 8),
```
martinoyovo, line 142: Since the space is only required when `arcGISImage` is not null, I think we should move the spacing inside the condition.
```
if (arcGISImage != null)
   Image.memory(arcGISImage.getEncodedBuffer()),
   const SizedBox(width: 8),
else
   const SizedBox.shrink(),
```
This ensures that the left spacing `(SizedBox(width: 8))` is only applied when `arcGISImage` is present. If `arcGISImage` is null, no extra space is added.

We can then eliminate the need for "spacing: 8".

And `SizedBox.shrink()` is preferred to be used as en empty widget than `Container()`.

changanxian: > And SizedBox.shrink() is preferred to be used as en empty widget than Container().

It is OK for this.

but I still feel better to update using `spacing` regardless the image is null or not, it still adds a little alignment for the legend name.

martinoyovo (approved):

PR #268, opened 2025-03-10
Cs/Add the mosaic rules sample

changanxian:
Add the Raster mosaic rule sample


lib/samples/apply_mosaic_rule_to_rasters/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## How to use the sample
+
+When the rasters are loaded, choose from a list of preset mosaic rules to apply to the rasters.
+
+## How it works
+
+1. Create an `ImageServiceRaster` using the service's URL.
+2. Create an `MosaicRule` object and set it to the `mosaicRule` property of the image service raster, if it does not specify a mosaic rule.
+3. Create an `RasterLayer` from the image service raster and add it to the map.
```
PaulAllanSturm, line 23: "Create an `MosaicRule`" should be "Create a `MosaicRule`" and "Create an `RasterLayer`" should be "Create a `RasterLayer`"

lib/samples/apply_mosaic_rule_to_rasters/apply_mosaic_rule_to_rasters.dart
```diff
@@ -0,0 +1,200 @@
...
+      ),
+      bottomSheet: _buildBottomSheet(),
+    );
+  }
+
+  Future<void> onMapViewReady() async {
+    // Create a map with a topographic basemap style.
+    final map = ArcGISMap.withBasemapStyle(BasemapStyle.arcGISTopographic);
+    _mapViewController.arcGISMap = map;
+    //Create a Raster with the provided URL
```
PaulAllanSturm, line 96: Some of the comments in this method should have a space after the `//` and a period at the end.

lib/samples/apply_mosaic_rule_to_rasters/apply_mosaic_rule_to_rasters.dart
```diff
@@ -0,0 +1,200 @@
...
+    await _mapViewController.setViewpointCenter(
+      rasterLayer.fullExtent!.center,
+      scale: 25000,
+    );
+    // Set the ready state variable to true to enable the sample UI.
+    setState(() => _ready = true);
+  }
+
+  // Update the mosaic method of the raster layer.
+  Future<void> _updateMosaicMethod() async {
```
PaulAllanSturm, line 119: This method doesn't use `await` anywhere, so it doesn't have to use the `async` keyword. And it can return simply `void`. And the calls to `setState(() => _ready = false)` and `setState(() => _ready = true)` won't actually do anything and are unnecessary.

lib/samples/apply_mosaic_rule_to_rasters/apply_mosaic_rule_to_rasters.dart
```diff
@@ -0,0 +1,200 @@
...
+    setState(() => _ready = false);
+    final rasterLayer =
+        _mapViewController.arcGISMap!.operationalLayers.firstWhere(
+              (layer) => layer is RasterLayer,
+            )
+            as RasterLayer;
+    final raster = rasterLayer.raster! as ImageServiceRaster;
```
PaulAllanSturm, line 126: We should just declare a member variable `late ImageServiceRaster _raster` on the class and then set it in `onMapViewReady`. That way we don't need to use `firstWhere` or use the `! as ImageSserviceRaster` cast.

lib/samples/apply_mosaic_rule_to_rasters/apply_mosaic_rule_to_rasters.dart
```diff
@@ -0,0 +1,200 @@
...
+    switch (_selectedMosaicMethod) {
+      case MosaicMethodEnum.objectID:
+        raster.mosaicRule = MosaicRule()..mosaicMethod = MosaicMethod.none;
```
PaulAllanSturm, line 129: I don't know if this will work, but can this be simplified to this:
```dart
raster.mosaicRule.mosaicMethod = MosaicMethod.none;
```

lib/samples/apply_mosaic_rule_to_rasters/apply_mosaic_rule_to_rasters.dart
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    setState(() => _ready = true);
+  }
+
+  // Build a bottom sheet to display mosaic method options.
+  Widget _buildBottomSheet() {
+    return BottomSheet(
+      onClosing: () {},
+      builder: (BuildContext context) {
```
PaulAllanSturm, line 148: `(BuildContext context)` can just be `(context)`

PR #270, opened 2025-03-20
Cs/Mosaic Rule Sample followup

changanxian:
- show a running spinner when the mosaic rule is changed(since it take a time)
 - follow the design to apply the different operation


harish-kunchala (approved):
They work great.

martinoyovo (approved):

PR #275, opened 2025-03-31
Readme and website links

harish-kunchala:


pubspec.yaml
```diff
@@ -16,11 +16,13 @@ dependencies:
   flutter:
```
jenmerritt, line 16: Any additional dependencies will need to be added to our internal doc

lib/widgets/sample_list_view.dart
```diff
@@ -42,6 +44,45 @@ class SampleListView extends StatelessWidget {
                 ),
               );
             },
+            trailing: PopupMenuButton(
+              icon: const Icon(Icons.more_vert),
```
jenmerritt, line 48: Visually, could the padding be more equal on the left and right hand sides? i.e. could we have less padding on the right. A bit more space can be given to the sample name and description and a bit less whitespace.
Screenshot 2025-04-01 at 12 33 11

lib/widgets/sample_list_view.dart
```diff
@@ -42,6 +44,45 @@ class SampleListView extends StatelessWidget {
...
+              itemBuilder:
+                  (context) => const [
+                    PopupMenuItem(
+                      value: 'README',
+                      child: ListTile(
+                        leading: Icon(Icons.description),
+                        title: Text('README'),
+                      ),
+                    ),
+                    PopupMenuItem(
```
jenmerritt, line 58: This may have been discussed and preferred by the team already, but can the developer website not just open in the browser? Not sure we need to embed it. I personally think an external link would be better. It's easier then to like grab the URL to save for later or bookmark etc. The link icon also says implies to me that is what will happen / was what I was expecting.

jenmerritt (commented):
Great to see this functionality coming to life.

One general comment, it would be useful for reviewers if you could provide a bit of a summary in the PR description of what you've implemented, and any comments you have.

lib/widgets/readme_page.dart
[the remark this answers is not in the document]
jenmerritt: The menu that is there right now is good. But there will basically need to be some kind of way of toggling between the readme / codeview / sample itself once a desired sample is selected. I think that's a key user workflow.

lib/widgets/sample_list_view.dart
PaulAllanSturm: It hasn't been discussed. I think I'd rather it pop out to the external browser too, like we do for the "Powered by Esri" link in the attribution bar -- `launchUrl(_esriUrl, mode: LaunchMode.externalApplication)`

lib/widgets/sample_list_view.dart
```diff
@@ -42,6 +44,45 @@ class SampleListView extends StatelessWidget {
...
+                      value: 'Website',
+                      child: ListTile(
+                        leading: Icon(Icons.link_rounded),
+                        title: Text('Website'),
+                      ),
+                    ),
+                  ],
+              onSelected: (String result) {
```
PaulAllanSturm, line 66: I think `(String result)` can just be `(result)` and the type will get properly inferred as `String`

lib/widgets/readme_page.dart
```diff
@@ -0,0 +1,100 @@
...
+      });
+    }
+  }
+
+  List<String> formatUrls() {
+    final keyUrls = <String>[];
+    final readmeUrl =
+        'https://raw.githubusercontent.com/Esri/arcgis-maps-sdk-flutter-samples/v.next/lib/samples/${widget.sample.key}/README.md';
+    final imageUrl =
+        'https://github.com/Esri/arcgis-maps-sdk-flutter-samples/raw/v.next/lib/samples/${widget.sample.key}/${widget.sample.key}.png';
```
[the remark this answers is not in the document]
PaulAllanSturm: My thought was we would bundle the READMEs into the app. So there wouldn't be a need to load them from over the network. We could use `build_runner` to process all the READMEs (and png's) into a form that can be loaded at runtime. (But this could be written up as a separate issue, too.)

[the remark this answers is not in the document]
jenmerritt: CC @kmueller-gis

lib/widgets/readme_page.dart
```diff
@@ -0,0 +1,100 @@
...
+
+  @override
+  void initState() {
+    super.initState();
+    _fetchMarkDown();
+  }
+
+  Future<void> _fetchMarkDown() async {
+    final readmeUrl = formatUrls()[0];
+    final imageUrl = formatUrls()[1];
```
PaulAllanSturm, line 28: These two lines are kind of weird. We create the same List twice and use magic numbers to index into it. Maybe just declare the two url's right here and do away with `formatUrls()`?

lib/widgets/sample_list_view.dart
harish-kunchala: Added Additional Padding of 20 dp to the left.

lib/widgets/readme_page.dart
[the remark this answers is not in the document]
kmueller-gis: @PaulAllanSturm or we could have that build_runner process copy all of the `.dart` and `README.md` files into an assets folder. Don't need the db file.

lib/widgets/sample_list_view.dart
harish-kunchala: Updated. Website now launches in an external browser.

lib/widgets/sample_list_view.dart
harish-kunchala: Updated.

lib/widgets/readme_page.dart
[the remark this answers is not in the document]
harish-kunchala: I've updated the sample to show AppBar options for README and Website.

lib/widgets/readme_page.dart
harish-kunchala: Rewmoved the function.

kmueller-gis (approved):

PaulAllanSturm (approved):

jenmerritt:
I will re-review. Either way, this PR should remain unmerged until after the release so as to not disrupt the release team.

pubspec.yaml
harish-kunchala: Done. Created [3123](https://devtopia.esri.com/runtime/flutter/issues/3123) with the following PR : [Added Markdown Package](https://devtopia.esri.com/runtime/flutter/pull/3125)

lib/widgets/readme_page.dart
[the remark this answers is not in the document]
harish-kunchala: Created Issue: 3127 to handle `build_runner` changes.

lib/widgets/sample_list_view.dart
jenmerritt: Looks much tidier thanks

lib/widgets/sample_list_view.dart
jenmerritt: Thanks - I think that works well.

lib/widgets/readme_page.dart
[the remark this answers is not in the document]
jenmerritt: The option in the app bar looks good 👍  

So right now you select the readme, it gets pushed in and so you use the app back button to return to the sample. I think this is good right now for the initial readme implementation - when we combine with the code view, we may want to play around with the navigation and experience of toggling between sample/readme/code view, but I see that as work to be done in either the code view PR or even a third PR that follows up after those. Hope that makes sense.
CC @kmueller-gis

jenmerritt (approved):
LGTM - as above, do not merge until after the release.

PR #277, opened 2025-04-02
Add loading indicator for Add raster from service sample

kmueller-gis:
Loading a raster layer from a service can take a long time. This PR adds a "Drawing..." popup to inform the user that the map is still working on rendering the layer.

**Summary of Changes**
- A new BusyIndicator widget was created and added to the common folder
- The BusyIndicator was applied to the Add raster from service sample

### DIFF
--- a/lib/common/busy_indicator.dart
+++ b/lib/common/busy_indicator.dart
@@ -1,0 +1,40 @@
+import 'package:flutter/material.dart';
+
+class BusyIndicator extends StatelessWidget {
+  const BusyIndicator({
+    required this.labelText,
+    required this.visible,
+    super.key,
+  });
+
+  final String labelText;
+  final bool visible;
+
+  @override
+  Widget build(BuildContext context) {
+    return Visibility(
+      visible: visible,
+      child: Center(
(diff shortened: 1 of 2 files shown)


harish-kunchala (approved):

lib/common/busy_indicator.dart
```diff
@@ -0,0 +1,44 @@
...
+  Widget build(BuildContext context) {
+    return Visibility(
+      visible: visible,
+      child: Center(
+        child: Container(
+          decoration: BoxDecoration(
+            color: Colors.white,
+            borderRadius: BorderRadius.circular(10),
+          ),
+          width: 110,
```
martinoyovo, line 27: I think we might not need to set the sizes here ?

kmueller-gis: The sizes are needed. Without them, the height runs full top to bottom, and the width is very tight.

martinoyovo: Clear now!

martinoyovo (approved):

PR #279, opened 2025-04-08
Add code view to sample viewer

kmueller-gis:
#### Summary of changes
- Added page to view code behind our samples
- Refactored the popup menu to a separate widget that is used in the sample list and detail page,


lib/widgets/code_view_page.dart
```diff
@@ -0,0 +1,123 @@
...
+class CodeViewPage extends StatefulWidget {
+  const CodeViewPage({required this.sample, super.key});
+  final Sample sample;
+
+  @override
+  State<StatefulWidget> createState() => _CodeViewPageState();
+}
+
+class _CodeViewPageState extends State<CodeViewPage> {
+  late final codeMap = _loadCodeMap(widget.sample.key);
```
PaulAllanSturm, line 15: I don't think this needs to be `late`

lib/widgets/code_view_page.dart
```diff
@@ -0,0 +1,123 @@
...
+  int selectedFileIndex = 0;
```
PaulAllanSturm, line 16: our convention is to do `var selectedFileIndex = 0;`

lib/widgets/code_view_page.dart
```diff
@@ -0,0 +1,123 @@
...
+        message:
+            'HTTP request failed with status: ${response.statusCode}: ${response.reasonPhrase}',
+      );
+    }
+
+    return response.body;
+  }
+}
+
+class CodeViewExeption implements Exception {
```
PaulAllanSturm, line 118: typo "CodeViewExeption" should be "CodeViewException" (missing "c")

lib/widgets/code_view_page.dart
```diff
@@ -0,0 +1,123 @@
...
+        child: Column(
+          children: [
+            Text(fileName, style: const TextStyle(fontWeight: FontWeight.bold)),
+            Expanded(
+              child: FutureBuilder(
+                future: codeMap[filePath],
+                builder: (
+                  BuildContext context,
+                  AsyncSnapshot<String> snapshot,
+                ) {
```
PaulAllanSturm, line 38: I think this could just be `builder: (context, snapshot) {` and let inference do the rest

lib/widgets/code_view_page.dart
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  Future<String> _fetchOnlineCodeForFile(String filePath) async {
+    const urlPrefix =
+        'https://raw.githubusercontent.com/Esri/arcgis-maps-sdk-flutter-samples/refs/heads/main/lib/samples/';
+    final fileUrl = '$urlPrefix/$filePath';
+    final fileUri = Uri.parse(fileUrl);
+    final response = await http.get(fileUri);
+    if (response.statusCode != 200) {
+      throw CodeViewExeption(
```
jenmerritt, line 107: This is throwing for me on the NMEA sample / 2 files.

jenmerritt (commented):
I think this has integrated well. I like how it integrates with the back button, but I think if it'd be useful to also have the menu in the AppBar on the code and readme pages. That way if I'm viewing the code and I can jump to the readme more easily / toggle between those pages without having to go back through the sample each time, for example.

lib/widgets/code_view_page.dart
jenmerritt: I'm sure this is what you were doing anyway but just to confirm, we will want to be able to view all of the dart files that appear in the metadata "snippets", whatever the number.

kmueller-gis: That is throwing because the NMEA code sample is not yet in the main branch. Once we merge the 200.7 branch with main, the NMEA code will show up.

lib/widgets/code_view_page.dart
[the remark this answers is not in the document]
kmueller-gis: Shouldn't these be added by GitHub?

lib/widgets/code_view_page.dart
jenmerritt: Facepalm - already tripped up over this! Could we display our message dialog saying like unable to fetch code data or something for now rather than throwing?

lib/widgets/code_view_page.dart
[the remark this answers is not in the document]
jenmerritt: I don't think we have the hook yet for samples - there is an issue for it.

lib/widgets/code_view_page.dart
jenmerritt: Is there a reason why the colors are there on the first file and not the second - is it something to do with the first file being a Widget and the second "just" a Class do you know? It's not the end of the world it just jumps out readability wise.

jenmerritt:
> I think if it'd be useful to also have the menu in the AppBar on the code and readme pages. That way if I'm viewing the code and I can jump to the readme more easily / toggle between those pages without having to go back through the sample each time, for example.

Quoting myself here.. feel free to write this up as an enhancement

PaulAllanSturm:
> > I think if it'd be useful to also have the menu in the AppBar on the code and readme pages. That way if I'm viewing the code and I can jump to the readme more easily / toggle between those pages without having to go back through the sample each time, for example.
> 
> Quoting myself here.. feel free to write this up as an enhancement

Yeah I'd rather do this as a followup issue. We'd have to restructure some things. We wouldn't want to do push/pop any more.

pubspec.yaml
```diff
@@ -18,6 +18,7 @@ dependencies:
   flutter_archive: ^6.0.3
   flutter_localizations:
     sdk: flutter
+  flutter_syntax_view: ^4.1.7
```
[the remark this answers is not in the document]
kmueller-gis: PR submitted

lib/widgets/code_view_page.dart
kmueller-gis: `late` is required here since a class property is being initialized by an instance function. Moved to the body of the build function removing the need for `late`.

lib/widgets/code_view_page.dart
kmueller-gis: Yep. Done.

lib/widgets/code_view_page.dart
kmueller-gis: Changed property to private. That allowed using `var` over `int`

lib/widgets/code_view_page.dart
kmueller-gis: good catch

lib/widgets/sample_info_popup_menu.dart
```diff
@@ -0,0 +1,72 @@
...
+              ),
+            );
+        }
+      },
+    );
+  }
+
+  Future<void> _launchSampleUrl() async {
+    if (!await launchUrl(webPageUrl(), mode: LaunchMode.externalApplication)) {
+      debugPrint('Could not launch ${webPageUrl()}');
```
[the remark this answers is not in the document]
kmueller-gis: The widget was changed to a StatefulWidget and `showMessageDialog()` replaced `debugPrint`

kmueller-gis:
> > > I think if it'd be useful to also have the menu in the AppBar on the code and readme pages. That way if I'm viewing the code and I can jump to the readme more easily / toggle between those pages without having to go back through the sample each time, for example.
> > 
> > 
> > Quoting myself here.. feel free to write this up as an enhancement
> 
> Yeah I'd rather do this as a followup issue. We'd have to restructure some things. We wouldn't want to do push/pop any more.

I've added a new issue to capture this thought. The issue includes Paul's push/pop concern.

PaulAllanSturm (approved):

jenmerritt:
Sounds good, just thinking about user experience.

jenmerritt (approved):

PR #282, opened 2025-04-11
Fixed app crash on cold start on Android due to an invalid layout size.

martinoyovo:


jenmerritt (approved):
I would just add a note in the merge commit that it was Android only.

PaulAllanSturm (approved):

PR #284, opened 2025-04-21
Enhancement/category page animations

martinoyovo:
**Animated Category + Ripple Search Transition**

This PR adds interactive animations to the category view and search navigation:

**Features**
- Animated entry for category cards (slide, fade, scale)
- Icon hover and tap effects for visual feedback
- Ripple transition from FAB to `SampleViewerPage` using a full-screen reveal
- Animated transition from category to sample list page

**How it Works**

- Category cards are built using a custom `CategoryCard` with internal animation controller
- Hint transitions rely on `SlideTransition`, `FadeTransition`, and `AnimatedSwitcher`
- Ripple effect uses `ClipPath` with a dynamic `CustomClipper` to mask the next page

**Preview**
https://github.com/user-attachments/assets/4ddc9e24-be2b-42fe-9b8f-757b16814d0f


jenmerritt:
Running in debug mode on my Android device I wasn't able to type in the search box and then the app crashed.

lib/utils/ripple_page_route.dart
```diff
@@ -0,0 +1,68 @@
...
+// limitations under the License.
+//
+
+import 'package:flutter/material.dart';
+
+class RipplePageRoute extends PageRouteBuilder {
+  RipplePageRoute({required this.position, required this.child})
+    : super(
+        transitionDuration: const Duration(milliseconds: 700),
+        pageBuilder: (_, __, ___) => child,
```
PaulAllanSturm, line 23: I think the unnamed parameters can now all be single underscores even when there is more than one parameter, so this could be `pageBuilder: (_, _, _) => child,`.

lib/main.dart
```diff
@@ -152,7 +164,8 @@ class _ResponsiveCategoryGrid extends StatelessWidget {
               child: CategoryCard(
                 category: SampleCategory.values[i],
                 onClick:
-                    () => _onCategoryClick(context, SampleCategory.values[i]),
+                    (d) => _onCategoryClick(context, SampleCategory.values[i]),
```
PaulAllanSturm, line 167: `d` is a weird parameter name. Should this just be `(_) =>`?

lib/widgets/category_card.dart
```diff
@@ -14,86 +14,237 @@
...
+  final int index;
+
+  @override
+  State<CategoryCard> createState() => _CategoryCardState();
+}
+
+class _CategoryCardState extends State<CategoryCard>
+    with SingleTickerProviderStateMixin {
+  double _scale = 1;
+  bool _iconPressed = false;
```
PaulAllanSturm, line 39: Normally we would say `var _scale = 1` and `var _iconPressed = false`, and let the type be inferred.

lib/widgets/category_card.dart
```diff
@@ -14,86 +14,237 @@
...
+  @override
+  void initState() {
+    super.initState();
+
+    _controller = AnimationController(
+      vsync: this,
+      duration: const Duration(milliseconds: 1000),
+    );
+
+    _slideAnimation = Tween<Offset>(
```
PaulAllanSturm, line 54: I think `Tween<Offset>(` could just be `Tween(` and let the type be inferred. Also the same for `Tween<double>(` below.

lib/widgets/category_card.dart
```diff
@@ -14,86 +14,237 @@
...
+  final bool pressed;
+  final int index;
+
+  @override
+  State<_AnimatedIcon> createState() => _AnimatedIconState();
+}
+
+class _AnimatedIconState extends State<_AnimatedIcon>
+    with SingleTickerProviderStateMixin {
+  bool _hovering = false;
```
PaulAllanSturm, line 167: This could be `var _hovering = false;`

lib/widgets/category_card.dart
martinoyovo: Removing the `Offset` type from `Tween` requires turning the begin/end value to double, which throws a lint error(`Unnecessary use of a 'double' literal.`). Now when I just remove the `Offset` type from `Tween`, I got an error: `A value of type 'Animation<int>' can't be assigned to a variable of type 'Animation<double>'`
Screenshot 2025-04-24 at 12 33 09 PM

PaulAllanSturm: ah okay, we can leave it as it is then

martinoyovo: Thanks @PaulAllanSturm!

PaulAllanSturm (approved):

PR #285, opened 2025-04-21
Enhancement/sample search animation

martinoyovo:
**Animated Search Experience in `SampleViewerPage`**

This PR introduces a dynamic **search experience** inside `SampleViewerPage`, improving both usability and visual quality for users navigating and discovering samples.

**Features**

  - Pulls from `Sample.title` and `Sample.keywords`
  - Filters out long phrases for visual clarity(max characters limit: 50)
  - Prefixed with randomized phrases like:`Try`, `Search`, `Explore`, `How about`, `Want to learn`, etc.
  - Adds `?` or `.` automatically for natural phrasing
  - Hint list refreshes every time the page is opened
  - Animated hint transitions integrated with `AnimatedSwitcher`

**How it Works**

- Uses a `Timer.periodic()` to change hint messages every 3 seconds  
- Ensures fallback if `_hintMessages` is empty to prevent crashes  

**Preview**
https://github.com/user-attachments/assets/f13df99b-8d77-49c7-8418-357e568f47e9

### DIFF
--- a/lib/widgets/sample_viewer_page.dart
+++ b/lib/widgets/sample_viewer_page.dart
@@ -75,21 +75,27 @@ class _SampleViewerPageState extends State<SampleViewerPage> {
   @override
   void initState() {
     super.initState();
-    loadSamples();
-    _searchFocusNode.addListener(() {
-      if (_searchFocusNode.hasFocus != _searchHasFocus) {
-        setState(() => _searchHasFocus = _searchFocusNode.hasFocus);
-      }
-    });
+    // Delay search after first frame.
+    WidgetsBinding.instance.addPostFrameCallback((_) async {
+      if (mounted) {
+        await loadSamples();
+        _searchFocusNode.addListener(() {
+          if (_searchFocusNode.hasFocus != _searchHasFocus) {
+            setState(() => _searchHasFocus = _searchFocusNode.hasFocus);
+          }
+        });
 
-    _hintTimer = Timer.periodic(const Duration(seconds: 3), (_) {
-      if (!mounted) return;
+        _hintTimer = Timer.periodic(const Duration(seconds: 3), (_) {
+          if (!mounted) return;
 
-      if (_hintMessages.isNotEmpty &&
-          !_searchHasFocus &&
-          _textEditingController.text.isEmpty) {
-        setState(() {
-          _currentHintIndex = (_currentHintIndex + 1) % _hintMessages.length;
+          if (_hintMessages.isNotEmpty &&
+              !_searchHasFocus &&
+              _textEditingController.text.isEmpty) {
+            setState(() {
+              _currentHintIndex =
+                  (_currentHintIndex + 1) % _hintMessages.length;
+            });
+          }
         });
       }
     });
--- a/lib/widgets/sample_viewer_page.dart
+++ b/lib/widgets/sample_viewer_page.dart
@@ -44,7 +44,7 @@ class _SampleViewerPageState extends State<SampleViewerPage> {
   var _ready = false;
   var _searchHasFocus = false;
 
-  final List<String> _searchPrefixes = [
+  final _searchPrefixes = [
     'Try',
     'Look for',
     'Search',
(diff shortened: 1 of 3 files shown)


lib/widgets/sample_viewer_page.dart
```diff
@@ -183,4 +240,44 @@ class _SampleViewerPageState extends State<SampleViewerPage> {
...
+        final useQuestion = random.nextBool();
+        final prefix = useQuestion
+            ? _questionPrefixes[random.nextInt(_questionPrefixes.length)]
+            : _searchPrefixes[random.nextInt(_searchPrefixes.length)];
+        final hint = useQuestion ? '$prefix "$title"?' : '$prefix "$title".';
+        uniqueHints.add(hint);
+      }
+
+      // Generate hints from short keywords.
+      for (final keyword in sample.keywords) {
```
jenmerritt, line 262: On one run of the app, it suggested I search for "field" which you could argue applies to all our samples in termsof fieldword, but the only one that came up was "find route in mobile map package". I realized after digging that it came out of "field mobility" as a sample keyword for that sample. 

There's an argument that the keywords could be improved in the designs themselves, but I do think you need to account for keywords not just being 1 word.

jenmerritt: Another example could be "find route" which has "shortest path" -> shortest and path alone wouldn't make as much sense.

martinoyovo:
> (posted on the wrong PR initially!)
> 
> Running in debug mode on my Android device I wasn't able to type in the search box and then the app crashed.
> 
> Works on emulator.

I couldn't reproduce this issue on my Pixel 9. Could you share the console output?

lib/widgets/sample_viewer_page.dart
martinoyovo: That's a good point. I'll try to update the code to see how it goes.

martinoyovo: I've added a condition to skip single-word keywords to avoir ambiguous suggestions. I believe that should fix the the issue.

jenmerritt: So say for Find Route, it includes:
- navigation
- shortest path

As long as both of those are covered, i.e. `navigation` is not skipped because it's one word.

lib/widgets/sample_viewer_page.dart
```diff
@@ -42,6 +44,35 @@ class _SampleViewerPageState extends State<SampleViewerPage> {
...
+    'Want to learn',
+    'How about',
+    "Let's explore",
+  ];
+
+  // Limit keywords 50 characters.
+  final int _maxHintLength = 50;
+
+  List<String> _hintMessages = [];
+  int _currentHintIndex = 0;
```
PaulAllanSturm, line 73: We want to use type inference as much as possible, so our pattern would be like this:
```dart
final _searchPrefixes = [ ... values ... ];
final _questionPrefixes = [ ... values ... ];
final _maxHintLength = 50;
final _hintMessages = <String>[];
var _currentHintIndex = 0;
```

lib/widgets/sample_viewer_page.dart
jenmerritt: I think if there isn't an effective regex that can be used to utilize the tags as they are in the readmes (as comma separated values) in order to provide meaningful suggestions then we shouldn't include them.

lib/widgets/sample_viewer_page.dart
```diff
@@ -42,6 +44,35 @@ class _SampleViewerPageState extends State<SampleViewerPage> {
...
+  final _searchPrefixes = [
+    'Try',
+    'Look for',
+    'Search',
+    'Explore',
+    'Type to explore',
+    'Check out',
+    'Discover',
+    'Start typing',
+    'Tap to search',
```
jenmerritt, line 56:
```suggestion
```
I think the "tap to search" instruction is a little confusing as it feels like the user should tap on the suggestion to search for it. The "Start typing" prefix in the list is a better expectation. The others look good too.

jenmerritt (approved):

jenmerritt:
I'm still seeing the crash with the latest change.

jenmerritt:
Ok the crash when you search happens on main/v.next if searching in debug mode. I'll write that up.

jenmerritt (approved):

jenmerritt:
LGTM - go ahead and merge

PR #286, opened 2025-04-22
Added Lazyloading

harish-kunchala:
For now our `ListView.builder` builds items on demand as they are scrolled into view. But as our sample data set grows, even the overhead of managing the list of items can be significant. 

Also this helps with our rendering issue, since only few items are loaded into our `Listview.builder` at a time.

Related to [#3248](https://devtopia.esri.com/runtime/flutter/issues/3248)


harish-kunchala:
Closing PR as the performance upgrades are non existent as we are not dealing with a huge dataset. The rendering seems to be resolved in the latest Flutter version.

PR #291, opened 2025-05-16
Change camera controller sample

kmueller-gis:
Implemented the "Change camera controller" sample.


lib/samples/change_camera_controller/change_camera_controller.png
jenmerritt, line 1: We should crop the `AppBar` out of the screengrab

lib/samples/change_camera_controller/change_camera_controller.dart
```diff
@@ -0,0 +1,233 @@
...
+      _selectedCameraControllerKind = cameraControllerKind;
+    });
+  }
+
+  ArcGISScene _setupScene() {
+    // Create a scene with an imagery basemap style.
+    final scene = ArcGISScene.withBasemapStyle(BasemapStyle.arcGISImagery);
+
+    // Set the scene's initial viewpoint.
+    scene.initialViewpoint = Viewpoint.withPointScaleCamera(
```
jenmerritt, line 174: This is more of an observation and thinking aloud - would become a separate issue as part of verifying the scenes API rather than a direct reflection on this sample PR. 

I went round in circles on the `Viewpoint` constructors that handle `Camera`. They seemed wordy but then actually required I think as there are a lot of ways of constructing a viewpoint. That said, are there are any logical opportunities to combine any constructors there?

lib/samples/change_camera_controller/change_camera_controller.dart
```diff
@@ -0,0 +1,233 @@
...
+        top: false,
+        left: false,
+        right: false,
+        child: Stack(
+          children: [
+            Column(
+              children: [
+                Expanded(
+                  // Add a scene view to the widget tree and set a controller.
+                  child: ArcGISSceneView(
```
jenmerritt, line 79: Per other comment, this observation would become a separate issue as part of verifying scenes rather than a direct reflection on this sample PR.

I've noticed that when a scene is loading, it takes quite a while for our interactions to kick in. Is this a reflection of the sample waiting to load or is this something we're doing in our interactions code in scene views? The delay feels longer than other SDKs is why I raise it. I can create a separate issue to test this more unless anyone had any initial thoughts.

jenmerritt (commented):
Looks good @kmueller-gis . Just a few thoughts.

lib/samples/change_camera_controller/change_camera_controller.dart
kmueller-gis: The gestures are prevented until the first time DrawStatus is complete. If that logic is too conservative, it will need to be revisited.

PaulAllanSturm: I think there might be a more local consideration here -- the `LoadingIndicator` widget doesn't get out of the way until after the plane graphic is downloaded and prepared. This sample doesn't really need a `LoadingIndicator` -- there aren't any problematic buttons or identify operations.

But even aside from that, definitely the "DrawStatus is complete" heuristic might be too conservative.

lib/samples/change_camera_controller/change_camera_controller.dart
jenmerritt: The verboseness is definitely helpful here, just making sure we revisit and review as we go. Thanks for your thoughts.

jenmerritt (approved):
LGTM from my initial comments

lib/samples/change_camera_controller/change_camera_controller.dart
PaulAllanSturm: I take back what I said about removing the LoadingIndicator -- we need `_planeGraphic` loaded before the user can select the `orbitGeoElement` camera, so we definitely need to disable user interaction until after `_planeGraphic` is loaded.

PaulAllanSturm (approved):

PR #298, opened 2025-05-30
cs/Add Web Scene from Portal item

changanxian:
Add a sample of loading a web scene from the portal item

### DIFF
--- a/lib/samples/display_web_scene_from_portal_item/display_web_scene_from_portal_item.dart
+++ b/lib/samples/display_web_scene_from_portal_item/display_web_scene_from_portal_item.dart
@@ -34,21 +34,11 @@ class _DisplayWebSceneFromPortalItemState
   @override
   Widget build(BuildContext context) {
     return Scaffold(
-      body: Stack(
-          children: [
-            Column(
-              children: [
-                Expanded(
-                  // Add the scene view to the widget tree and set a controller.
-                  child: ArcGISSceneView(
-                    controllerProvider: () => _sceneViewController,
-                    onSceneViewReady: onSceneViewReady,
-                  ),
-                ),
-              ],
-            ),
-          ],
-        ),
+      // Add the scene view to the body of the scaffold and set a controller.
+      body: ArcGISSceneView(
+        controllerProvider: () => _sceneViewController,
+        onSceneViewReady: onSceneViewReady,
+      ),
     );
   }
 


lib/samples/display_web_scene_from_portal_item/display_web_scene_from_portal_item.dart
```diff
@@ -0,0 +1,76 @@
...
+    with SampleStateSupport {
+  // Create a controller for the scene view.
+  final _sceneViewController = ArcGISSceneView.createController();
+  // A flag for when the scene view is ready and controls can be used.
+  var _ready = false;
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      body: SafeArea(
```
PaulAllanSturm, line 39: We don't need a SafeArea because there are no controls at the bottom. We want edge-to-edge.

lib/samples/display_web_scene_from_portal_item/display_web_scene_from_portal_item.dart
```diff
@@ -0,0 +1,76 @@
...
+                  // Add the scene view to the widget tree and set a controller.
+                  child: ArcGISSceneView(
+                    controllerProvider: () => _sceneViewController,
+                    onSceneViewReady: onSceneViewReady,
+                  ),
+                ),
+              ],
+            ),
+            // Display a progress indicator and prevent interaction until state is ready.
+            LoadingIndicator(visible: !_ready),
```
PaulAllanSturm, line 57: We don't need a LoadingIndicator because there are no controls (or onTap handlers) to be disabled.

lib/samples/display_web_scene_from_portal_item/display_web_scene_from_portal_item.dart
```diff
@@ -0,0 +1,65 @@
...
+  final _sceneViewController = ArcGISSceneView.createController();
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      body: Stack(
+          children: [
+            Column(
+              children: [
+                Expanded(
```
PaulAllanSturm, line 41: We don't need the Stack, Column, or Expanded. When there is only the ArcGISSceneView, the widget tree can be as simple as:
```dart
    return Scaffold(
      body: ArcGISSceneView(
        controllerProvider: () => _sceneViewController,
        onSceneViewReady: onSceneViewReady,
      ),
    );
```

PR #299, opened 2025-06-02
cs/Add a point scene layer

changanxian:
- Add a scene point layer.


lib/samples/add_point_scene_layer/add_point_scene_layer.dart
```diff
@@ -0,0 +1,97 @@
...
+  @override
+  void dispose() {
+    super.dispose();
+    _loadStatusSubscription.cancel();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      body: SafeArea(
```
PaulAllanSturm, line 47: SafeArea isn't necessary because this sample doesn't have controls. We want edge-to-edge.

lib/samples/add_point_scene_layer/add_point_scene_layer.dart
```diff
@@ -0,0 +1,97 @@
...
+            ),
+            // Display a progress indicator and prevent interaction until state is ready.
+            LoadingIndicator(visible: !_ready),
+          ],
+        ),
+      ),
+    );
+  }
+
+  void onSceneViewReady() {
```
PaulAllanSturm, line 72: There should be more tutorial comments in this method.

lib/samples/add_point_scene_layer/add_point_scene_layer.dart
```diff
@@ -0,0 +1,97 @@
...
+      Uri.parse(
+        'https://tiles.arcgis.com/tiles/V6ZHFr6zdgNZuVG0/arcgis/rest/services/Airports_PointSceneLayer/SceneServer/layers/0',
+      ),
+    );
+    scene.operationalLayers.add(pointSceneLayer);
+    _loadStatusSubscription = pointSceneLayer.onLoadStatusChanged.listen((status) {
+      if (status == LoadStatus.loaded) {
+        setState(() => _ready = true);
+      }
+    });
```
PaulAllanSturm, line 95: This could more simply be:
```dart
pointSceneLayer.load().then((_) {
});
```

But I'm not sure why this sample needs a `LoadingIndicator`. There aren't any controls (or onTap handlers) that we want to prevent from being used.

changanxian: `_ready` is used to control the visibility of the loading progress indicator. Since the sample is to load a mass point scene layer, it will take a few seconds to load, it is good to have the progress spinner running until the layer is loaded.

changanxian:
```
pointSceneLayer.load().then((_) {
});
```
We like to set the `_ready` to be true until the layer is loaded, not just begin to load.

lib/samples/add_point_scene_layer/add_point_scene_layer.dart
changanxian: fixed

lib/samples/add_point_scene_layer/add_point_scene_layer.dart
PaulAllanSturm: > ```
> pointSceneLayer.load().then((_) {
>   setState(() => _ready = true);
> });
> ```
> 
> We like to set the `_ready` to be true until the layer is loaded, not just begin to load.

That's what this code does. Once the `Future` returned by `load()` completes, the `then()` callback runs.

PaulAllanSturm (approved):

PR #303, opened 2025-06-06
New sample: Display scene from mobile scene package

jenmerritt:
- This PR adds the new sample "display scene from mobile scene package"
- Data is downloaded from AGO. Note that I added some "downloading data" text since it takes a while to download. We have an issue to make this more consistent across all the samples.
- Tested on iOS and Android and reported observations in the issue.


lib/samples/display_scene_from_mobile_scene_package/display_scene_from_mobile_scene_package.dart
```diff
@@ -0,0 +1,86 @@
+//
+// Copyright 2024 Esri
```
PaulAllanSturm, line 2: 2025

jenmerritt: Spot the copy paste

PaulAllanSturm (approved):

PR #312, opened 2025-06-27
cs/Update internal download utility

changanxian:
By switching to the internal HttpClient, there is a slight performance gain compared to using an external HttpClient created manually here. The change results in about a 10–15% performance improvement, but downloading is still slow compared to other platforms.


jenmerritt (approved):
Please be specific in the issue with which samples should be tested for verification to avoid confusion (i.e. we haven't updated all samples that involve data yet)

PR #314, opened 2025-07-01
Style point with scene symbol

martinoyovo:


lib/samples/style_point_with_scene_symbol/style_point_with_scene_symbol.dart
```diff
@@ -0,0 +1,130 @@
...
+  final _sceneViewController = ArcGISSceneView.createController();
+  // A flag for when the scene view is ready and controls can be used.
+  var _ready = false;
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      body: Stack(
+        children: [
+          ArcGISSceneView(
```
jenmerritt, line 40:
```suggestion
          // Add a scene view to the widget tree and set a controller.
          ArcGISSceneView(
```
Let's try to catch these things that are the same in every sample.

jenmerritt (approved):

PR #316, opened 2025-07-02
ShowViewshed from GeoElement Sample

harish-kunchala:
Created ShowView from GeoElement Sample.


lib/samples/show_viewshed_from_geoelement_in_scene/README.md
```diff
@@ -0,0 +1,48 @@
...
+* ModelSceneSymbol
+* OrbitGeoElementCameraController
+
+## Offline data
+
+Offline sample data will be downloaded by the sample viewer automatically.
+
+| Link | Local Location |
+|------|----------------|
+|[Model Marker Symbol Data](https://www.arcgis.com/home/item.html?id=07d62a792ab6496d9b772a24efea45d0)| `<userhome>`/ArcGIS/Runtime/Data/3D/bradley_low_3ds/bradle.3ds |
```
PaulAllanSturm, line 40: This "Local Location" path doesn't make sense on Mobile -- it only makes sense for the other SDKs that support Desktop. When we've been given one of these tables in other Samples, we remove the table but leave the hyperlink. So like this:
```txt
## Offline data

[Model Marker Symbol Data](https://www.arcgis.com/home/item.html?id=07d62a792ab6496d9b772a24efea45d0)
```

I'm looking at some other samples, and we're not really consistent when there's both an "Offline data" section and a "About the data" section. So I'm not sure what's best. At minimum we don't present the `<userhome>/....` path, which then means there's no need for a single-cell markdown table.

lib/samples/show_viewshed_from_geoelement_in_scene/show_viewshed_from_geoelement_in_scene.dart
```diff
@@ -0,0 +1,288 @@
...
+class _ShowViewshedFromGeoelementInSceneState
+    extends State<ShowViewshedFromGeoelementInScene>
+    with SampleStateSupport {
+  // Create a controller for the scene view.
+  final _sceneViewController = ArcGISSceneView.createController();
+
+  // The graphic for the tank.
+  Graphic? _tankGraphic;
+
+  // Timer for animation
```
PaulAllanSturm, line 40: period at the end

lib/samples/show_viewshed_from_geoelement_in_scene/show_viewshed_from_geoelement_in_scene.dart
```diff
@@ -0,0 +1,288 @@
...
+        child: Stack(
+          children: [
+            ArcGISSceneView(
+              controllerProvider: () => _sceneViewController,
+              onSceneViewReady: onSceneViewReady,
+              onTap: onTap,
+            ),
+            // Display a progress indicator and prevent interaction until state is ready.
+            LoadingIndicator(visible: !_ready),
+            // Banner at the top
```
PaulAllanSturm, line 65: period

PaulAllanSturm: (here and various other places below)

lib/samples/show_viewshed_from_geoelement_in_scene/show_viewshed_from_geoelement_in_scene.dart
```diff
@@ -0,0 +1,288 @@
...
+      Uri.parse(
+        'https://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/Buildings_Brest/SceneServer/layers/0',
+      ),
+    );
+    scene.operationalLayers.add(buildingsLayer);
+
+    return scene;
+  }
+
+  // Handle the on tap function.
```
PaulAllanSturm, line 131: This comment could be more specific. Maybe like "Convert the tapped location to a waypoint in the scene and begin the tank animation to the waypoint."

lib/samples/show_viewshed_from_geoelement_in_scene/show_viewshed_from_geoelement_in_scene.dart
```diff
@@ -0,0 +1,288 @@
...
+    // Return the graphic that combines geometry and symbol.
+    return Graphic(
+      geometry: ArcGISPoint(x: -4.506390, y: 48.385624),
+      attributes: {'HEADING': 0.0},
+      symbol: tankSymbol,
+    );
+  }
+
+  // Adds the tank graphic to a graphics overlay and sets the initial viewpoint.
+  Future<void> _addTankToScene(Graphic tankGraphic) async {
```
PaulAllanSturm, line 239: There's no `await` in this function, so it can be declared more simply `void _addTankToScene(Graphic tankGraphic) {`. And then it doesn't need to be `await`ed where it is called above.

lib/samples/show_viewshed_from_geoelement_in_scene/show_viewshed_from_geoelement_in_scene.dart
```diff
@@ -0,0 +1,288 @@
...
+      ..sceneProperties.headingExpression = '[HEADING]'
+      ..sceneProperties.pitchExpression = '[PITCH]'
+      ..sceneProperties.rollExpression = '[ROLL]';
+    graphicsOverlay.renderer = renderer;
+
+    _sceneViewController.graphicsOverlays.add(graphicsOverlay);
+  }
+
+  // Add viewshed to the scene
+  Future<void> _addViewShedToScene(Graphic tankGraphic) async {
```
PaulAllanSturm, line 258: This method also doesn't need `Future` and `async`

lib/samples/show_viewshed_from_geoelement_in_scene/show_viewshed_from_geoelement_in_scene.dart
```diff
@@ -0,0 +1,288 @@
...
+    // Create a GeoElementViewShed attached to the scene.
+    final geoElementViewShed =
```
PaulAllanSturm, line 260: In the comment and in the variable name, we should standardize the capitalization to use a lower case "s" in "Viewshed". (It's spelled "Viewshed" in our API.)

lib/samples/show_viewshed_from_geoelement_in_scene/README.md
harish-kunchala: Yeah. Maybe we should make a note on tackling *Offline Data* segments in our Samples Notes.

lib/samples/show_viewshed_from_geoelement_in_scene/show_viewshed_from_geoelement_in_scene.dart
harish-kunchala: done.

lib/samples/show_viewshed_from_geoelement_in_scene/show_viewshed_from_geoelement_in_scene.dart
```diff
@@ -0,0 +1,288 @@
...
+      ),
+    );
+    scene.operationalLayers.add(buildingsLayer);
+
+    return scene;
+  }
+
+  // Convert the tapped location into a waypoint within the scene and initiate the tank's animation towards the waypoint.
+  Future<void> onTap(Offset localPosition) async {
+    // Convert localPosition to scenePoint,
```
PaulAllanSturm, line 133: er well now it has a comma, not a period

lib/samples/show_viewshed_from_geoelement_in_scene/show_viewshed_from_geoelement_in_scene.dart
```diff
@@ -0,0 +1,288 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      body: SafeArea(
+        top: false,
+        left: false,
+        right: false,
+        child: Stack(
+          children: [
+            ArcGISSceneView(
```
PaulAllanSturm, line 58: It's customary to say "`// Add a scene view to the widget tree and set a controller.`" above this line

lib/samples/show_viewshed_from_geoelement_in_scene/show_viewshed_from_geoelement_in_scene.dart
```diff
@@ -0,0 +1,288 @@
...
+      ..sceneProperties.pitchExpression = '[PITCH]'
+      ..sceneProperties.rollExpression = '[ROLL]';
+    graphicsOverlay.renderer = renderer;
+
+    _sceneViewController.graphicsOverlays.add(graphicsOverlay);
+  }
+
+  // Add viewshed to the scene.
+  void _addViewshedToScene(Graphic tankGraphic) {
+    // Create a GeoElementViewShed attached to the scene.
```
PaulAllanSturm, line 259: The "S" should be lowercase "s" in "GeoElementViewShed"

harish-kunchala: Forgot to push the change. done.

PR #320, opened 2025-07-08
Real-time progress bar updates in mission settings modal

harish-kunchala:
The modal bottom sheet was using `StatefulBuilder` but wasn't being notified when the parent widget's `_progress` value changed. The `_onTick` method was calling `setState()` on the main widget, but the modal sheet remained static until manually refreshed.

- Added modal state tracking with `_isMissionModalOpen` flag and `_modalStateSetter` reference
- Modified `_onTick` method to update the modal sheet state when animation is playing and modal is open

### DIFF
--- a/lib/samples/animate_3d_graphic/animate_3d_graphic.dart
+++ b/lib/samples/animate_3d_graphic/animate_3d_graphic.dart
@@ -38,6 +38,10 @@ class _Animate3dGraphicState extends State<Animate3dGraphic>
   // The graphic for the plane.
   Graphic? _planeGraphic;
 
+  // Track mission modal status.
+  bool _isMissionModalOpen = false;
+  StateSetter? _modalStateSetter;
+
   // The camera controller that follows the plane.
   late OrbitGeoElementCameraController _cameraController;
 
@@ -76,7 +80,7 @@ class _Animate3dGraphicState extends State<Animate3dGraphic>
   // The current progress of the animation.
   double get _progress {
     if (_frames.isEmpty) return 0;
-    return _currentFrameIndex / _frames.length;
+    return _currentFrameIndex.toDouble() / _frames.length.toDouble();
   }
 
   @override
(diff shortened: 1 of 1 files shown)


lib/samples/animate_3d_graphic/animate_3d_graphic.dart
```diff
@@ -380,6 +384,11 @@ class _Animate3dGraphicState extends State<Animate3dGraphic>
         _isPlaying = false;
       }
     });
+
+    // Update modal sheet if it's open
```
PaulAllanSturm, line 388: period at the end of the comment

lib/samples/animate_3d_graphic/animate_3d_graphic.dart
```diff
@@ -380,6 +384,11 @@ class _Animate3dGraphicState extends State<Animate3dGraphic>
...
+    if (_isMissionModalOpen && _modalStateSetter != null) {
+      _modalStateSetter!(() {});
+    }
```
PaulAllanSturm, line 391: It seems that `_isMissionModalOpen` is unnecessary and can just be deleted everywhere. It doesn't affect anything that isn't already taken care of by `_modalStateSetter`.

Also, I think you can just use `?` instead of a null check followed by `!`.
```dart
_modalStateSetter?(() {});
```

lib/samples/animate_3d_graphic/animate_3d_graphic.dart
harish-kunchala: done.

lib/samples/animate_3d_graphic/animate_3d_graphic.dart
harish-kunchala: done, also added wrapped `StatefulBuilder` around progress text and `LinearProgressIndicator` as only those two widgets have to rebuild for every frame.

PaulAllanSturm (approved):

PR #327, opened 2025-07-16
cs/Using gesture detector 

changanxian:
- Using GestureDetector to decide if one of the GeoViews needs to update the other one. It is more reliable than using the navigation point changed callback.

### DIFF
--- a/lib/samples/match_viewpoint_of_geo_views/match_viewpoint_of_geo_views.dart
+++ b/lib/samples/match_viewpoint_of_geo_views/match_viewpoint_of_geo_views.dart
@@ -85,7 +85,7 @@ class _MatchViewpointOfGeoViewsState extends State<MatchViewpointOfGeoViews>
             _isMapViewInteraction = true;
             _isSceneViewInteraction = false;
           }),
-          onDoubleTapDown: (details) => setState(() {
+          onDoubleTapDown: (_) => setState(() {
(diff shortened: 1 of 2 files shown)


lib/samples/match_viewpoint_of_geo_views/match_viewpoint_of_geo_views.dart
```diff
@@ -83,16 +79,39 @@ class _MatchViewpointOfGeoViewsState extends State<MatchViewpointOfGeoViews>
...
-        child: ArcGISMapView(
-          controllerProvider: () => _mapViewController,
-          onMapViewReady: onMapViewReady,
+        child: GestureDetector(
+          behavior: HitTestBehavior.translucent,
+          onTapDown: (_) => setState(() {
+            _isMapViewInteraction = true;
+            _isSceneViewInteraction = false;
+          }),
+          onDoubleTapDown: (details) => setState(() {
```
PaulAllanSturm, line 88: For consistency, `(details)` should be `(_)`. Also the same for the ArcGISSceneView's onTapDown and onDoubleTapDown callbacks.

PaulAllanSturm:
The `README.md` explicitly says to "Wire up the ... `onNavigationChanged` event". Since we're not doing that, the README.md should be updated to match the implementation.

PaulAllanSturm (approved):

PR #332, opened 2025-07-18
Create the image frames just once before the animation starts

PaulAllanSturm:
Create all the image frames just once when the sample loads, then simply assign each one to the ImageOverlay with each animation step. With this approach, the large memory allocations happen only once, which prevents the garbage collector from getting overwhelmed.

### DIFF
--- a/lib/samples/animate_images_with_image_overlay/animate_images_with_image_overlay.dart
+++ b/lib/samples/animate_images_with_image_overlay/animate_images_with_image_overlay.dart
@@ -49,8 +49,8 @@ class _AnimateImagesWithImageOverlayState
   var _opacity = 0.5;
   // Create an ImageOverlay to display the animated images.
   final _imageOverlay = ImageOverlay();
-  // A list to hold the image URIs for the animation.
-  List<Uri> _imageUriList = [];
+  // A list to hold the image frames for the animation.
+  List<ImageFrame?> _imageFrames = [];
   // A string to display the download progress.
   var _downloadProgress = '';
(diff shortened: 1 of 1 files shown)


changanxian:
This is the same as the initial implementation. Since it kept crashing in fast mode within seconds, we switched to using an image list (in the Swift sample) or a file path list (in the QT sample) to update images.

Retested it again — it crashes on my iPhone SE in less than 10 seconds (release mode, sometimes it crashes less in debug mode).

PaulAllanSturm:
On my iPhone 14, this version runs for 20 minutes and stays at a constant 2.5 GB of memory usage (as shown in the Xcode memory tab).

In the version where a new ArcGISImage is created for every animation tick, the memory usage grows steadily until the app is killed for using too much memory.

Is it possible your iPhone SE can't handle 2.5 GB of memory usage?

harish-kunchala:
The only problem I see is if there are lots of images, they stay in the memory all the time. Instead of this can we consider using any cache or other lazy loading techniques ?

PaulAllanSturm:
The fundamental problem is that we can't trigger the garbage collector to release the memory used by an image frame. That limitation prevents us from successfully doing any sort of cache/lazy-loading techniques. Regardless of the approach, the animation will load those frames in rapid sequence, which will consume too much memory too quickly.

A solution would require us to provide a method call that can release the memory associated with an ArcGISImage (basically a `dispose()` method like dart:ui's Image class). That would free us from having to rely on the garbage collector -- we can instead free the memory immediately when we switch from one frame to the next.

PR #334, opened 2025-07-28
Replace MediaQuery.of(context)

PaulAllanSturm:
`MediaQuery.of(context)` is inefficient (your widget will rebuild for every change to any property of `MediaQuery`). We should always use a more specific accessor, such as `MediaQuery.sizeOf(context)`.


martinoyovo (approved):
A great catch, well done!

PR #349, opened 2025-08-21
Calculate the current frame based on the animation's elapsed time

PaulAllanSturm:
The `Ticker` object provides an `elapsed` parameter that tells us how much time has elapsed since the animation began. We should use this to calculate which animation frame to use, so that the animation isn't dependent on the frame rate (i.e., we don't want a device running at 120 fps to finish twice as fast as a device running at 60 fps).

This required changing how we were starting and stopping the ticker -- before the ticker was running continuously, while now the ticker only runs while the animation is running. This ensures that the `elapsed` parameter reports the time since the animation started, rather than since the sample opened.

### DIFF
--- a/lib/samples/animate_3d_graphic/animate_3d_graphic.dart
+++ b/lib/samples/animate_3d_graphic/animate_3d_graphic.dart
@@ -86,8 +86,8 @@ class _Animate3dGraphicState extends State<Animate3dGraphic>
   @override
   void initState() {
     super.initState();
-    // Start the animation ticker.
-    _ticker = createTicker(_onTick)..start();
+    // Create the animation ticker.
+    _ticker = createTicker(_onTick);
   }
 
   @override
@@ -334,10 +334,13 @@ class _Animate3dGraphicState extends State<Animate3dGraphic>
       }
     }
 
+    if (_ticker.isActive) {
+      _ticker.stop();
+      setState(() => _isPlaying = false);
+    }
     setState(() {
       _frames = frames;
       _currentFrameIndex = 0;
-      _isPlaying = false;
     });
 
     if (_frames.isNotEmpty) {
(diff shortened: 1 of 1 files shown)


martinoyovo:
In this sample, we're currently using the `TickerProviderStateMixin`, which is designed for scenarios involving multiple tickers or animation controllers. However, since our implementation only requires a single ticker, we should use the `SingleTickerProviderStateMixin` instead. It's more optimized for single ticker use cases and has less overhead.

martinoyovo (commented):
While reviewing the code, I noticed some areas of improvement which aren't related the current edits but didn't know where to address them so I'll put them in the comments:

**Area 1:**
We are currently using the `TickerProviderStateMixin`, which is designed for scenarios involving multiple tickers or animation controllers. However, since our implementation only requires a single ticker, we should use the `SingleTickerProviderStateMixin` instead.

**Area 2:**
The current `_updateFrame()` trigger full widget rebuilds at 60fps, causing unnecessary UI recomputation:
```
setState(() {
  _altitude = frame.position.z ?? 0;
  _heading = frame.heading;
  _pitch = frame.pitch;
  _roll = frame.roll;
});
```
If we use a `ValueNotifier` for telemetry data  class and `ValueListenableBuilder` for the telemetry display widget, that could improve the animation performance:

Something like this:
```
class TelemetryData {
  final ValueNotifier<double> altitude = ValueNotifier(0);
  final ValueNotifier<double> heading = ValueNotifier(0);
  final ValueNotifier<double> pitch = ValueNotifier(0);
  final ValueNotifier<double> roll = ValueNotifier(0);
  
  void dispose() {
    altitude.dispose();
    heading.dispose();
    pitch.dispose();
    roll.dispose();
  }
}

void _updateFrame(Frame frame) {
  if (_planeGraphic == null) return;

  _planeGraphic!.geometry = frame.position;
  _planeGraphic!.attributes['HEADING'] = frame.heading;
  _planeGraphic!.attributes['PITCH'] = frame.pitch;
  _planeGraphic!.attributes['ROLL'] = frame.roll;

  // Update telemetry without setState
  _telemetryData.altitude.value = frame.position.z ?? 0;
  _telemetryData.heading.value = frame.heading;
  _telemetryData.pitch.value = frame.pitch;
  _telemetryData.roll.value = frame.roll;
}
```

**Area 3:**
The code below runs 60 times per second in the animation loop, which performs a floating-point division on every simple animation frame:
```
final framesPerSecond = 60 * _animationSpeed.frameStep;
var nextFrame = elapsed.inMilliseconds ~/ (1000 / framesPerSecond);
```
The problem is that the `framesPerSecond` value doesn't change during animation, but we recalculate it every frame. We should only calculate it when the animation speed changes. So I propose we add a calculation helper:

```
void _updateFrameInterval() {
  final framesPerSecond = 60 * _animationSpeed.frameStep;
  _frameIntervalMs = (1000 / framesPerSecond).round();
}
```
that way, we can call during `initState` and when `animation speed changes`.
We could also add a guard in _onTick to exit early if there's no data:
`if (_frames.isEmpty) return;`

changanxian (approved):

PR #356, opened 2025-09-08
Cs/Unifying the sample data downloading expriences

changanxian:
- Centralize the sample data download and management experience on a single page


lib/samples/display_scene_from_mobile_scene_package/display_scene_from_mobile_scene_package.dart
```diff
@@ -49,48 +47,15 @@ class _DisplaySceneFromMobileScenePackageState
             controllerProvider: () => _sceneViewController,
             onSceneViewReady: onSceneViewReady,
           ),
```
PaulAllanSturm, line 49: Now that the download indicator is gone, the `_ready` flag is no longer used and the `Stack` widget isn't necessary. So we can delete those and simplify the widget tree.

lib/samples/find_route_in_mobile_map_package/README.metadata.json
```diff
@@ -1,6 +1,6 @@
 {
     "category": "Maps",
-    "description": "Display maps and use locators to enable search and routing offline using a Mobile Map Package.",
+    "description": "Display maps and use locators to enable search and routing offline using a Mobile Map Package.",
```
PaulAllanSturm, line 3: this change isn't necessary

lib/samples/find_route_in_mobile_map_package/find_route_in_mobile_map_package.dart
```diff
@@ -71,7 +57,7 @@ class _FindRouteInMobileMapPackageState
       body: SafeArea(
         // Display a list of maps from the mobile map packages once loaded.
         child: FutureBuilder(
-          future: mobileMapPackages,
+          future: _loadMobileMapPackages(),
```
PaulAllanSturm, line 60: This causes a problem with FutureBuilder -- every time the `build()` method is called, `_loadMobileMapPackages()` will be called, which will create a new Future each time. It is important to create the Future only once. Usually it would be created `initState()`.

changanxian:
> I think this will cause problems for `lib/utils/sample_runner.dart`. Samples launched from sample_runner.dart won't be able to fetch the `extras` from `GoRouter`. I think we should address that as a separate issue.

Create a separate issue to trace 4019

lib/widgets/downloadable_resources_page.dart
```diff
@@ -0,0 +1,293 @@
...
+                onPressed: _isDownloading
+                    ? _cancelDownload
+                    : _isComplete
+                    ? _openSample
+                    : _startDownload,
+                child: Text(
+                  _isDownloading
+                      ? 'Cancel'
+                      : _isComplete
+                      ? 'Open'
```
jenmerritt, line 138: I guess we don't need this anymore? On Android I saw the 'open' button display briefly before the sample then loaded.

lib/widgets/downloadable_resources_page.dart
```diff
@@ -0,0 +1,293 @@
...
+          icon: const Icon(Icons.arrow_back),
+          onPressed: () {
+            if (_isDownloading) {
+              unawaited(_cancelDownload());
+            }
+            context.pop();
+          },
+        ),
+      ),
+      body: Padding(
```
jenmerritt, line 84: When the data is already downloaded and you're re-opening the sample again from the menu, the downloads content is briefly shown before the sample is loaded, like a flash. I'm wondering if the check whether it's downloaded shouldn't be at a higher level? Rather than displaying the page, checking, and then opening the sample.

@martinoyovo could you maybe help test and weigh in on this?

jenmerritt: Or does this related to issue 4019?

PR #430, opened 2026-01-28
New sample: query dynamic entities

martinoyovo:


lib/samples/query_dynamic_entities/query_dynamic_entities.dart
```diff
@@ -0,0 +1,605 @@
...
+            controller: controller,
+            decoration: const InputDecoration(
+              hintText: 'Flight_396',
+              labelText: 'Flight Number',
+            ),
+            autofocus: true,
+          ),
+          actions: [
+            TextButton(
+              onPressed: () => Navigator.pop(context),
```
PaulAllanSturm: This is a small thing, but tapping the "Cancel" button dismisses both this AlertDialog and also the QueryMenu. I think "Cancel" should only dismiss the AlertDialog, but leave the QueryMenu up.

PaulAllanSturm (commented):
I'm about half way through the main file. I'll do the rest in another batch.

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
```diff
@@ -0,0 +1,605 @@
...
+      _PlaneAttributeKey.altitudeFeet ||
+      _PlaneAttributeKey.speed => FieldType.float64,
+      _ => FieldType.text,
+    };
+  }
+}
+
+// A custom dynamic entity data provider that streams mock air traffic data for PHX.
+final class _PhxAirTrafficProvider extends CustomDynamicEntityDataProvider {
+  _PhxAirTrafficProvider({this.localJsonPath});
```
PaulAllanSturm: Add a comment above this constructor.

PaulAllanSturm (commented):
finished reviewing

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
martinoyovo: Fixed. I've removed  the `navigator.pop `from "With Flight Number" `QueryMenu` Item and it resolved the issue.

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
```diff
@@ -0,0 +1,605 @@
...
+            autofocus: true,
+          ),
+          actions: [
+            TextButton(
+              onPressed: () => Navigator.pop(context),
+              child: const Text('Cancel'),
+            ),
+            FilledButton(
+              onPressed: () => Navigator.pop(context, controller.text),
+              child: const Text('Done'),
```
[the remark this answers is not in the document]
martinoyovo: That's interesting because it's working on my end and no exception was thrown locally.

https://github.com/user-attachments/assets/db2cbe5e-e966-4b0a-9661-e53824795e8e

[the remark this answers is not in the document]
martinoyovo: The initial text "Flight_396" that you see is the text field's placeholder. You'd need to input the Flight number before you can see results. I'll update the placeholder so it won't feel like the text field has a default value.

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
```diff
@@ -0,0 +1,596 @@
+// Copyright 2026 Esri
```
jenmerritt, line 2:
```suggestion
//
// Copyright 2026 Esri
```

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
```diff
@@ -0,0 +1,605 @@
...
+    final phxPoint = ArcGISPoint(
+      x: -112.0101,
+      y: 33.4352,
+      spatialReference: SpatialReference.wgs84,
+    );
+    _phoenixAirportBuffer = GeometryEngine.bufferGeodetic(
+      geometry: phxPoint,
+      distance: 15,
+      distanceUnit: LinearUnit(unitId: LinearUnitId.miles),
+      maxDeviation: double.nan,
```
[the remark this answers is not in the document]
jenmerritt: Interesting thought if this is commonly going to be set - though not sure it's part of the API design or other implementations from what I can see? Could be something to ask in `geometry-questions` before we look to deviate. Maybe there's background. Unless I'm missing something and this is an auditing mishap, in which case definitely write up!

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
```diff
@@ -0,0 +1,596 @@
...
+
+    // Load and configure the dynamic entity layer.
+    _loadDynamicEntityLayer();
+    _configureDynamicEntityLayer(_dynamicEntityLayer);
+    map.operationalLayers.add(_dynamicEntityLayer);
+
+    // Set the map to the map view.
+    _mapViewController.arcGISMap = map;
+
+    setState(() => _ready = true);
```
jenmerritt, line 145:
```suggestion
    // Set the ready state variable to true to enable the sample UI.
    setState(() => _ready = true);
```

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
```diff
@@ -0,0 +1,596 @@
...
+
+  // Load the dynamic entity layer with a custom data source.
+  void _loadDynamicEntityLayer() {
+    final listPaths = GoRouter.of(context).state.extra! as List<String>;
+
+    // Create a custom data source that streams PHX air traffic observations.
+    final provider = _PhxAirTrafficProvider(localJsonPath: listPaths.first);
+    _dataSource = CustomDynamicEntityDataSource(provider);
+
+    _dynamicEntityLayer = DynamicEntityLayer(_dataSource);
```
jenmerritt:
```suggestion
   // Create a dynamic entity later from the custom data source.
    _dynamicEntityLayer = DynamicEntityLayer(_dataSource);
```
This may feel verbose - but generally where it's API very specific to the sample use case (so here dynamic entities) it's best to make sure we're explicit with comments.

lib/samples/query_dynamic_entities/README.md
```diff
@@ -0,0 +1,61 @@
...
+attribute criteria. The query returns a collection of dynamic entities matching the
+`DynamicEntityQueryParameters` or track IDs at the moment the query is executed.
+
+An example of this is a flight tracking app that monitors airspace near a particular airport,
+allowing the user to monitor flights based on different criteria such as arrival airport or flight
+number.
+
+## How to use the sample
+
+Tap **Query Flights** button and choose one of the query options. After the query completes, a list of matching flights is shown. Expand a flight to see its latest attributes update in real-time.
```
jenmerritt:
```suggestion
Tap the "Query Flights" button and select a query to perform from the menu. Once the query is complete, a list of the resulting flights will be displayed. Tap on a flight to see its latest attributes in real-time.
```
Per the design

lib/samples/query_dynamic_entities/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## How it works
+
+1. Create a `DynamicEntityDataSource` to stream dynamic entity events.
+2. Create `DynamicEntityQueryParameters` and set its properties:
+   * To spatially filter results, set `geometry` and `spatialRelationship`.
+   * To query entities with certain attribute values, set `whereClause`.
```
jenmerritt, line 26: I think we're missing the third bullet point here from the design:
`* To get entities with specific track IDs, modify the trackIDs collection.`

Unless we're deliberately not doing that in the implementation.

lib/samples/query_dynamic_entities/README.md
```diff
@@ -0,0 +1,61 @@
...
+* DynamicEntityDataSource
+* DynamicEntityLayer
+* DynamicEntityObservation
+* DynamicEntityQueryParameters
+* DynamicEntityQueryResult
+
+## Offline data
+
+This sample uses [PHX Air Traffic JSON](https://www.arcgis.com/home/item.html?id=c78e297e99ad4572a48cdcd0b54bed30).
+It is downloaded from ArcGIS Online as part of the sample.
```
jenmerritt: Do we need this part? I don't see it in the design.

jenmerritt (commented):
User testing looks good. A few additional comments on doc/readme from me.

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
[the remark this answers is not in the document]
PaulAllanSturm: Yeah, Swift and Kotlin do not provide a default value. Our normal audit process for determining default values is to compare a method's "short form" override to its "long form" override, and figuring out what default values the "short form" uses. This particular method doesn't have a "short form" and a "long form", so there was no audit mishap here. I wouldn't recommend we go through the prose remarks of every parameter in the API, trying to infer some intended default value. So thus and therefore let's leave this as-is.

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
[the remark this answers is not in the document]
PaulAllanSturm: Oh! Right. Yes, I was confused by the placeholder.

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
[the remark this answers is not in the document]
martinoyovo: Alright, no update here then

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
```diff
@@ -0,0 +1,605 @@
...
+                                padding: EdgeInsets.all(16),
+                                child: Text(
+                                  'There are no flights to display for this query.',
+                                  textAlign: TextAlign.center,
+                                ),
+                              ),
+                            );
+                          }
+
+                          return ListView.builder(
```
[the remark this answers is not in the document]
martinoyovo: Added comments

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
martinoyovo: Due to prior changes, the commit suggestion is outdated but I've included the comment in the branch update. Thanks for the feedback.

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
```diff
@@ -0,0 +1,605 @@
...
+
+    // Show the query results in a bottom sheet.
+    await _showQueryResultsSheet(selection, entities: entities);
+  }
+
+  Future<void> _showQueryResultsSheet(
+    QuerySelection selection, {
+    required List<DynamicEntity> entities,
+  }) async {
+    await showModalBottomSheet<void>(
```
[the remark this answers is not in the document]
martinoyovo: Done! But I didn't use the default `BottomSheetSettings` here because we need the results list to be in a constrained height that we can scroll through.

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
```diff
@@ -0,0 +1,648 @@
...
+          ..whereClause = "status = 'In flight' AND arrival_airport = 'PHX'";
+        queryResult = await dataSource.queryDynamicEntities(params);
+      case QueryType.trackId:
+        queryResult = await dataSource.queryDynamicEntitiesByTrackIds([
+          selection.trackId!,
+        ]);
+    }
+
+    // Collect the query results into a list.
+    entities = queryResult.iterator().toList(growable: false);
```
PaulAllanSturm: I think this should just be:
```dart
final entities = queryResult.iterator().toList(growable: false);
```
I don't see a reason why `entities` is declared so much earlier in the method as a `var` with an empty list.

lib/samples/query_dynamic_entities/query_dynamic_entities.dart
```diff
@@ -0,0 +1,648 @@
...
+
+  // Run the selected query and display the results in a bottom sheet.
+  Future<void> _runQuery(QuerySelection selection) async {
+    final dataSource = _dataSource;
+    if (dataSource == null) return;
+    // Show the buffer graphic only for geometry queries.
+    _bufferGraphicsOverlay.isVisible = selection.type == QueryType.geometry;
+
+    // A list to collect the query results.
+    var entities = const <DynamicEntity>[];
```
PaulAllanSturm: this is weird

martinoyovo: I was essentially matching the format with `DynamicEntityQueryResult queryResult`; definition, but I'll remove that.

jenmerritt (approved):

PaulAllanSturm (approved):

PR #438, opened 2026-02-18
New Configure Scene Environment sample

kmueller-gis:
Implementation of the [Configure scene environment](https://devtopia.esri.com/runtime/common-samples/tree/main/designs/ConfigureSceneEnvironment) sample.

Sample that illustrates how a developer can modify the environment settings for a scene. This currently only applies to local scenes but will also affect global scenes when they are merged into LocalSceneView. 

The scene environment controls are contained in a bottom sheet vs a fixed area at the bottom of the app window. As this is a new sample, any suggestions are welcome.


lib/samples/configure_scene_environment/configure_scene_environment.dart
```diff
@@ -0,0 +1,396 @@
...
+  const ConfigureSceneEnvironment({super.key});
+
+  @override
+  State<ConfigureSceneEnvironment> createState() =>
+      _ConfigureSceneEnvironmentState();
+}
+
+class _ConfigureSceneEnvironmentState extends State<ConfigureSceneEnvironment>
+    with SampleStateSupport {
+  // Get a controller for the ArcGISLocalSceneView
```
PaulAllanSturm: period!

lib/samples/configure_scene_environment/configure_scene_environment.dart
```diff
@@ -0,0 +1,396 @@
...
+  final _localSceneViewController = ArcGISLocalSceneView.createController();
+
+  // Flag to activate the settings bottom sheet.
+  bool _showBottomSheet = false;
```
PaulAllanSturm: `var` instead of `bool`

lib/samples/configure_scene_environment/configure_scene_environment.dart
```diff
@@ -0,0 +1,396 @@
...
+              children: [
+                Expanded(
+                  // Add a local scene view to the widget tree and set a controller.
+                  child: ArcGISLocalSceneView(
+                    controllerProvider: () => _localSceneViewController,
+                    onLocalSceneViewReady: onLocalSceneReady,
+                  ),
+                ),
+                Center(
+                  // Button to summon the scene environment settings.
```
PaulAllanSturm: "summon" is a little disconcerting

lib/samples/configure_scene_environment/configure_scene_environment.dart
```diff
@@ -0,0 +1,396 @@
...
+            )
+          : null,
+    );
+  }
+
+  Future<void> onLocalSceneReady() async {
+    // Create and load the local scene from a ArcGISOnline web scene.
+    final websceneUri = Uri.parse(
+      'https://maps.arcgis.com/home/item.html?id=fcebd77958634ac3874bbc0e6b0677a4',
+    ); // Local scene with 3D trees and buildings
```
PaulAllanSturm: This comment tucked in down here should just be part of the comment up above

lib/samples/configure_scene_environment/configure_scene_environment.dart
```diff
@@ -0,0 +1,396 @@
...
+  final ArcGISLocalSceneViewController localSceneViewController;
+
+  @override
+  State<StatefulWidget> createState() => _SceneEnvironmentSettingsState();
+}
+
+class _SceneEnvironmentSettingsState extends State<SceneEnvironmentSettings> {
+  // Listing used for background color drop down.
+  final _backgroundColorOptions = [
+    (name: 'None', color: const Color.fromARGB(0, 0, 0, 0)),
```
PaulAllanSturm: Does `Colors.transparent` work for 'None'?

lib/samples/configure_scene_environment/configure_scene_environment.dart
```diff
@@ -0,0 +1,396 @@
...
+    (name: 'White', color: Colors.white),
+  ];
+
+  // Initialize state variables for the control panel.
+  var _isStarsEnabled = true;
+  var _isAtmosphereEnabled = true;
+  var _isDirectShadowsEnabled = false;
+  var _backgroundColor = Colors.black;
+  var _isSunLighting = false;
+  var _lightingDateTime = DateTime.utc(2026, 3, 20, 12);
```
PaulAllanSturm, line 151: What happens on March 20?

lib/samples/configure_scene_environment/configure_scene_environment.dart
```diff
@@ -0,0 +1,396 @@
...
+
+      // Record the time zone offset if one was set on the web scene.
+      if (sunLighting.displayTimeZone != null) {
+        _lightingTimeZoneOffset = Duration(
+          hours: sunLighting.displayTimeZone!.hours,
+          minutes: sunLighting.displayTimeZone!.minutes,
+        );
+      }
+
+      _lightingHour = sunLighting.simulatedDate
```
PaulAllanSturm, line 185: Comment above this line? something like  `// Record the current lighting hour on the web scene.`

lib/samples/configure_scene_environment/configure_scene_environment.dart
```diff
@@ -0,0 +1,396 @@
...
+        const Divider(),
+        // Section for the background color setting controls.
+        Row(
+          children: [
+            const Text('Background color:'),
+            const Spacer(),
+            DropdownButton(
+              value: _backgroundColor,
+              items: _backgroundColorOptions
+                  .map<DropdownMenuItem<Color>>(
```
PaulAllanSturm: I think `<DropdownMenuItem<Color>>` is unnecessary

lib/samples/configure_scene_environment/README.metadata.json
```diff
@@ -0,0 +1,33 @@
+{
+    "category": "Scenes",
```
jenmerritt, line 2: Not relevant to this review but it's just occurred to me we could consider a LS category and/or relevant tags. I've raised this in the samples design council channel. If we decide to do something I'll create PRs on all the local scene sample designs and we can amend.

lib/samples/configure_scene_environment/configure_scene_environment.dart
```diff
@@ -0,0 +1,396 @@
...
+
+class _ConfigureSceneEnvironmentState extends State<ConfigureSceneEnvironment>
+    with SampleStateSupport {
+  // Get a controller for the ArcGISLocalSceneView
+  final _localSceneViewController = ArcGISLocalSceneView.createController();
+
+  // Flag to activate the settings bottom sheet.
+  bool _showBottomSheet = false;
+
+  // A flag for when the map view is ready and controls can be used.
```
jenmerritt:
```suggestion
  // A flag for when the local scene view is ready and controls can be used.
```

lib/samples/configure_scene_environment/configure_scene_environment.dart
```diff
@@ -0,0 +1,396 @@
...
+    final scene = ArcGISScene.withUri(websceneUri)!;
+    await scene.load();
+
+    // Set the scene on the local scene view.
+    _localSceneViewController.arcGISScene = scene;
+
+    setState(() {
+      // The view is ready for interaction.
+      _ready = true;
+    });
```
jenmerritt: Can this be one line?
```suggestion
// The view is ready for interaction.
setState(() => _ready = true);
```

lib/samples/configure_scene_environment/configure_scene_environment.dart
```diff
@@ -0,0 +1,396 @@
...
+}
+
+// Widget containing all of the controls for configuring the scene environment settings.
+class SceneEnvironmentSettings extends StatefulWidget {
+  const SceneEnvironmentSettings({
+    required this.localSceneViewController,
+    super.key,
+  });
+
+  final ArcGISLocalSceneViewController localSceneViewController;
```
jenmerritt: This is more a question out of curiosity and for my education. CC @PaulAllanSturm too.

Are there scenarios where passing the controller around is more risky than others? I take it here when we know that bottom sheet is tied to our main sample where we know the local scene view lifecycle is tied it's safe. But I guess if we had some other usage of this widget and the local scene was to be disposed then this wouldn't work?

lib/samples/configure_scene_environment/configure_scene_environment.dart
```diff
@@ -0,0 +1,396 @@
...
+  var _isAtmosphereEnabled = true;
+  var _isDirectShadowsEnabled = false;
+  var _backgroundColor = Colors.black;
+  var _isSunLighting = false;
+  var _lightingDateTime = DateTime.utc(2026, 3, 20, 12);
+  var _lightingTimeZoneOffset = Duration.zero;
+  var _lightingHour = 12;
+
+  // Convenience variable to get the ArcGISScene from the controller.
+  ArcGISScene get _scene => widget.localSceneViewController.arcGISScene!;
```
jenmerritt: Are we using the controller in here other than to get the scene? Could we pass the scene? 

Again probably more a curiosity thing than any specific suggestion.

lib/samples/configure_scene_environment/configure_scene_environment.dart
```diff
@@ -0,0 +1,396 @@
...
+      );
+    } else {
+      newSceneLighting = VirtualLighting(
+        areDirectShadowsEnabled: _isDirectShadowsEnabled,
+      );
+    }
+
+    // Set the new lighting object to the scene.
+    widget.localSceneViewController.arcGISScene?.environment.lighting =
+        newSceneLighting;
```
jenmerritt: Can we just use the scene here? Do we need to get it from the controller again?

lib/samples/configure_scene_environment/configure_scene_environment.dart
kmueller-gis: Spring Equinox. Wanted to default to an even amount of day and night time.

lib/samples/configure_scene_environment/README.metadata.json
kmueller-gis: Eventually, LocalSceneView and SceneView will be one in the same. The distinction will be between local and global scenes at that point. Keep this in mind when advocating for new categories.

jenmerritt: Yeah that's what we've just been discussing. I'm personally worried about customers navigating this. But I agree I think I'd get pushback proposing this for the reasons you state.

lib/samples/configure_scene_environment/configure_scene_environment.dart
PaulAllanSturm: yeah this should just use the `_scene` getter, like all the rest

lib/samples/configure_scene_environment/configure_scene_environment.dart
PaulAllanSturm: Probably it makes more sense just to pass in the scene. That makes the dependency more explicit. This widget wouldn't have to worry about `arcGISScene` being null. And this widget doesn't actually need to interact with the controller (other than to fetch the scene).

lib/samples/configure_scene_environment/configure_scene_environment.dart
PaulAllanSturm: That's a reasonable concern. Here the controller is a required parameter to the constructor, which should be sufficient to instruct the coder that they shouldn't invalidate it out from under this widget.

It would probably be instructive to do a spike of a more complex app built with say, BLoC, and explore what it means for your ArcGIS SDK objects to live in the middle "business logic" layer. I suspect it would be best to have the controller owned by the state widget that owns the MapView/SceneView. They just have such a tight 1:1 lifecycle correspondence. Then all your operational layers, feature tables, route tasks, etc would be owned by some "model" object, and when your MapView/SceneView is loaded, it would populate itself from the model.

PaulAllanSturm: (though in this specific scenario, the dependency should be narrowed down to just the scene anyway)

lib/samples/configure_scene_environment/configure_scene_environment.dart
PaulAllanSturm: Well that's just very intelligent. I figured it was your wife's birthday or something.

lib/samples/configure_scene_environment/configure_scene_environment.dart
kmueller-gis: Updated the bottom sheet to take a scene rather than a controller.

lib/samples/configure_scene_environment/configure_scene_environment.dart
kmueller-gis: Controller has been removed from this widget.

lib/samples/configure_scene_environment/configure_scene_environment.dart
kmueller-gis: done

lib/samples/configure_scene_environment/configure_scene_environment.dart
kmueller-gis: done

lib/samples/configure_scene_environment/configure_scene_environment.dart
kmueller-gis: I guess I shouldn't be listening to demon hunting podcasts while writing comments. It is a beast of a dialog box though...

lib/samples/configure_scene_environment/configure_scene_environment.dart
kmueller-gis: The comment doesn't really add anything. It has been removed.

lib/samples/configure_scene_environment/configure_scene_environment.dart
kmueller-gis: Yep. Changed.

lib/samples/configure_scene_environment/configure_scene_environment.dart
kmueller-gis: A comment has been added.

lib/samples/configure_scene_environment/configure_scene_environment.dart
kmueller-gis: Correct. Removed.

lib/samples/configure_scene_environment/configure_scene_environment.dart
kmueller-gis: fixed

lib/samples/configure_scene_environment/configure_scene_environment.dart
kmueller-gis: yes

PaulAllanSturm (approved):

lib/samples/configure_scene_environment/configure_scene_environment.dart
jenmerritt: Sounds good - I'll capture this.

lib/samples/configure_scene_environment/configure_scene_environment.dart
jenmerritt: Sounds good!

jenmerritt (approved):

PR #444, opened 2026-03-03
Update "Filter Features in Scene" Sample

jenmerritt:
The purpose of this PR is to replace the deprecated OSM Buildings Layer. As part of the change, we utilize a new 3D basemap and so there are some tweaks to the configuration of the scene and some simplification to the workflow of the sample, including applying the filter before the detailed buildings, to better demonstrate the filter capability.

Summary of changes:
- Replace deprecated OSM data with new 3D navigation basemap
- Update the screengrab since the attribution has changed
- Make the relevant updates to the readme and metadata
- Restructure the configuration of the Scene
- Just show/hide the data rather than re-loading
- Reorder the steps so we "filter" first, and then show the detailed buildings, then reset.
- Clarified some of the comments to make the distinction clear between the "detailed" buildings layer vs the 3D buildings base layer that is being filtered.

| Before | After |
|---------|---------|
| step1-before |  step1-after | 
|step2-before | step2-after|
| step3-before | step3-after |


harish-kunchala (approved):

jenmerritt:
@changanxian nudge incase this has been missed

changanxian (approved):

PR #449, opened 2026-03-11
Cs/Pin the TTS to 4.2.2

changanxian:
- Pin the flutter_tts to 4.2.2.


jenmerritt:
We don't want to do this on `main`. This branch is designed to work with 200.8 plus the minimum supported version at the time of release i.e. 3.22.x. So we need to keep the versions low and instead look to pin dependencies as needed, where there are build issues.

PaulAllanSturm (approved):

jenmerritt (approved):

PR #450, opened 2026-03-11
Update the TTS to 4.2.5

changanxian:


jenmerritt:
This is fine to update - but I added a comment to the issue, the priority is to fix the build break on main by pinning flutter_tts to a lower version there so that users are able to use `main` with `200.8` and the [minimum version of Flutter it supports](https://developers.arcgis.com/flutter/system-requirements/system-requirements-for-200-8/) (3.22.x)

jenmerritt (approved):

PaulAllanSturm (approved):

PR #451, opened 2026-03-11
cs/Update using new callout

changanxian:
- Update reverse geocoding sample to callout showing the address.


lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode.dart
```diff
@@ -130,6 +130,10 @@ class _FindAddressWithReverseGeocodeState
         .where((str) => str.isNotEmpty)
         .join(', ');
 
-    showMessageDialog(combinedString);
+    _mapViewController.callout.showAt(
+      normalizedTapPoint,
+      detail: combinedString,
+      leaderPosition: .top,
```
jenmerritt: My only comment would be, I quite like the Callout in the position like in Swift below which is how it tends to be presented - so above the icon:

Image

I'll also let @martinoyovo give his thoughts / preference as well though.

jenmerritt (commented):
Nice to see! 👏

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode.dart
changanxian: Yes, I like the style also.

jenmerritt (approved):

martinoyovo (approved):

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode.dart
martinoyovo: The leader position was set to `.top`, that is why the callout was showing at the bottom of the pin icon. It should be `.bottom` is we want the callout "always" to be on top.

LeaderPosition.top -> Callout is shown at the bottom
LeaderPosition.bottom -> Callout is shown at the top

PR #453, opened 2026-03-17
New sample: Apply Map Algebra

jenmerritt:
- Implements the Apply Map Algebra sample
- Verified on:
   - iOS device ✅ 
   - Android device  ✅


jenmerritt:
@Rachael-E would you mind doing a first pass? Just to help verify the implementation in terms of the sample design. I'll then assign the team to delve into the UI implementation etc. Thanks!

lib/samples/apply_map_algebra/apply_map_algebra.dart
```diff
@@ -0,0 +1,323 @@
...
+    // Category: Raised shore line areas.
+    final isRaisedShoreline = tenMeterBinField
+        .isGreaterThanOrEqualToValue(0)
+        .logicalAnd(tenMeterBinField.isLessThanValue(10));
+
+    // Category: Ice covered areas.
+    // Note: Operator overloads are available on some operators (e.g. /, *, +, -) to allow for more concise syntax when
+    // chaining operations. Instead of using logicalAnd, & can also be used to combine boolean fields, as shown below.
+    final isIceCovered =
+        tenMeterBinField.isGreaterThanOrEqualToValue(10) &
```
Rachael-E, line 238: This could be updated if and when the SDK is able to support overloads for greater than etc

Rachael-E (approved):
Looks great! Just a comment on the operator overloads which is very much an if and when thing.

lib/samples/apply_map_algebra/apply_map_algebra.dart
jenmerritt: Sounds good - since this is an enhancement it's currently a stretch goal for this release and may not make it in

lib/samples/apply_map_algebra/apply_map_algebra.dart
```diff
@@ -0,0 +1,323 @@
...
+  // Raster layers to represent the original elevation raster and the results of the analysis.
+  RasterLayer? _originalElevationRasterLayer;
+  RasterLayer? _resultsRasterLayer;
+
+  // A variable to manage the active raster layer.
+  RasterLayer? _selectedRasterLayer;
+
+  // Variables to manage the UI.
+  bool _isPerformingAnalysis = false;
+  bool _hasAnalysisResults = false;
```
PaulAllanSturm: these should be `var` (like `var _ready` below)

lib/samples/apply_map_algebra/apply_map_algebra.dart
```diff
@@ -0,0 +1,323 @@
...
+                          children: [
+                            Expanded(
+                              // Sets the original elevation raster layer as active.
+                              child: ListTile(
+                                leading: Icon(
+                                  _selectedRasterLayer ==
+                                          _originalElevationRasterLayer
+                                      ? Icons.radio_button_checked
+                                      : Icons.radio_button_unchecked,
+                                  color: _ready ? null : Colors.grey,
```
PaulAllanSturm: I don't think `_ready ? null` is doing anything useful (here and the other radio button below). These controls are inside a `Visibility(visible: _hasAnalysisResults`, and `_hasAnalysisResults` doesn't get set to true until for sure `_ready` is already true. I think this should just be
```dart
color: Colors.grey
```

lib/samples/apply_map_algebra/apply_map_algebra.dart
```diff
@@ -0,0 +1,323 @@
...
+      scale: 500000,
+    );
+
+    // Create a raster from a raster elevation file.
+    final raster = Raster.withFileUri(_elevationFile.uri);
+    // Create a raster layer from the raster.
+    _originalElevationRasterLayer = RasterLayer.withRaster(raster);
+
+    // Create a stretch renderer to visualize the elevation raster layer using the surface preset color ramp.
+    // Set up stretch parameters with min and max values of elevation data
```
PaulAllanSturm: period at the end?

lib/samples/apply_map_algebra/apply_map_algebra.dart
```diff
@@ -0,0 +1,323 @@
...
+      longitude: -5.234218,
+      scale: 500000,
+    );
+
+    // Create a raster from a raster elevation file.
+    final raster = Raster.withFileUri(_elevationFile.uri);
+    // Create a raster layer from the raster.
+    _originalElevationRasterLayer = RasterLayer.withRaster(raster);
+
+    // Create a stretch renderer to visualize the elevation raster layer using the surface preset color ramp.
```
PaulAllanSturm, line 168: I'd maybe add a blank line after here, otherwise visually it looks like this comment is tied to only the MinMaxStretchParameters bit, when really it is summarizing the next three steps.

lib/samples/apply_map_algebra/apply_map_algebra.dart
```diff
@@ -0,0 +1,323 @@
...
+      600,
+    );
+
+    // Assign values to the geomorphic categories and evaluate.
+    // Raised shoreline=1, ice covered=2, ice-free high ground=3.
+    final geomorphicCategoryField = await tenMeterBinField
+        .replaceWithValueIf(selection: isRaisedShoreline, value: 1)
+        .replaceWithValueIf(selection: isIceCovered, value: 2)
+        .replaceWithValueIf(selection: isIceFreeHighGround, value: 3)
+        .evaluate();
```
PaulAllanSturm, line 253: Maybe `evaluate()` should be a separate step. Give it a little more emphasis.

lib/samples/apply_map_algebra/apply_map_algebra.dart
```diff
@@ -0,0 +1,323 @@
...
+      if (exportedFiles.isEmpty) {
+        setState(() => _isPerformingAnalysis = false);
+        return;
+      }
+
+      // Use the exported raster file to create a new raster with the geomorphic categorization.
+      final geomorphicCategorizationRaster = Raster.withFileUri(
+        exportedFiles.first,
+      );
+      // Use the raster to create a raster layer for the displaying the results on the map.
```
PaulAllanSturm: One too many "the" at "for the displaying"

lib/samples/apply_map_algebra/apply_map_algebra.dart
```diff
@@ -0,0 +1,323 @@
...
+      _resultsRasterLayer = RasterLayer.withRaster(
+        geomorphicCategorizationRaster,
+      );
+
+      // Define a renderer for the different geomorphic categories.
+      final colors = <int, Color>{
```
PaulAllanSturm: The `<int, Color>` bit might be unnecessary.

PaulAllanSturm (approved):

martinoyovo (approved):

PR #458, opened 2026-03-30
Harish/dynamic entity layer

harish-kunchala:
- Added Callout display for `add_dynamic_entity_layer.dart` sample.


lib/samples/add_dynamic_entity_layer/README.md
```diff
@@ -0,0 +1,47 @@
+# Add dynamic entity layer
+
+Display data from an ArcGIS stream service using a dynamic entity layer.
+
+![Image of display dynamic entity layer](add_dynamic_entity_layer.png)
```
jenmerritt:
```suggestion
![Image of add dynamic entity layer](add_dynamic_entity_layer.png)
```

lib/samples/add_dynamic_entity_layer/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Use case
+
+A stream service is a type of service provided by ArcGIS Velocity and GeoEvent Server that allows clients to receive a stream of data observations via a web socket. ArcGIS Maps SDK for .NET allows you to connect to a stream service and manage the information as dynamic entities and display them in a dynamic entity layer. Displaying information from feeds such as a stream service is important in applications like dashboards where users need to visualize and track updates of real-world objects in real-time.
```
jenmerritt:
```suggestion
A stream service is a type of service provided by ArcGIS Velocity and GeoEvent Server that allows clients to receive a stream of data observations via a web socket. ArcGIS Maps SDK for Flutter allows you to connect to a stream service and manage the information as dynamic entities and display them in a dynamic entity layer. Displaying information from feeds such as a stream service is important in applications like dashboards where users need to visualize and track updates of real-world objects in real-time.
```

lib/samples/add_dynamic_entity_layer/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## How to use the sample
+
+Use the controls to connect to or disconnect from the stream service, modify display properties in the dynamic entity layer, and purge all observations from the application.
+
+## How it works
+
+1. Create an `ArcGIStreamService` using a `Uri`.
+2. Set a `DynamicEntityFilter` on the stream service to limit the amount of data coming from the server.
+3. Set the `MaximumDuration` property of the stream service `PurgeOptions` to limit the amount of data managed by the application.
```
jenmerritt:
```suggestion
3. Set the `maximumDuration` property of the stream service `purgeOptions` to limit the amount of data managed by the application.
```

lib/samples/add_dynamic_entity_layer/README.md
```diff
@@ -0,0 +1,47 @@
...
+4. Create a `DynamicEntityLayer` using the stream service.
+5. Update values in the layer's `TrackDisplayProperties` to customize the layer's appearance.
+6. Add the `DynamicEntityLayer` to the map.
+7. Configure a `onTap` event handler on the `MapView` to select a dynamic entity and display the entity's attributes in a callout.
```
jenmerritt:
```suggestion
7. Configure an `onTap` event handler on the `ArcGISMapView` to select a dynamic entity and display the entity's attributes in a callout.
```

lib/samples/add_dynamic_entity_layer/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Relevant API
+
+* ArcGISStreamService
+* ConnectionStatus
+* DynamicEntity
+* DynamicEntityFilter
+* DynamicEntityLayer
+* DynamicEntityPurgeOptions
```
jenmerritt:
```suggestion
* DynamicEntityDataSourcePurgeOptions
```

lib/samples/add_dynamic_entity_layer/README.md
```diff
@@ -0,0 +1,47 @@
...
+5. Update values in the layer's `TrackDisplayProperties` to customize the layer's appearance.
+6. Add the `DynamicEntityLayer` to the map.
+7. Configure a `onTap` event handler on the `MapView` to select a dynamic entity and display the entity's attributes in a callout.
+
+## Relevant API
+
+* ArcGISStreamService
+* ConnectionStatus
+* DynamicEntity
+* DynamicEntityFilter
```
jenmerritt:
```suggestion
* ArcGISStreamServiceFilter
```

lib/samples/add_dynamic_entity_layer/README.md
```diff
@@ -0,0 +1,47 @@
...
+* DynamicEntityPurgeOptions
+* TrackDisplayProperties
+
+## About the data
+
+This sample uses a [stream service](https://realtimegis2016.esri.com:6443/arcgis/rest/services/SandyVehicles/StreamServer) that simulates live data coming from snowplows near Sandy, Utah. There are multiple vehicle types and multiple agencies operating the snowplows.
+
+## Additional information
+
+More information about dynamic entities can be found in the [guide documentation](https://developers.arcgis.com/net/real-time/work-with-dynamic-entities/).
```
jenmerritt:
```suggestion
More information about dynamic entities can be found in the [guide documentation](https://developers.arcgis.com/flutter/real-time/work-with-dynamic-entities/).
```
This URL is broken right now, but the page exists on the `next` branch of the documentation site (the link to this is in the quick links in Slack, for future reference). When the 300 samples go to main, this URL will be correct.

lib/samples/add_dynamic_entity_layer/README.metadata.json
jenmerritt, line 1: Note this will need to updated per the class changes in the readme

jenmerritt (commented):
Readme comments:

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+
+  // A flag for when the map view is ready and controls can be used.
+  var _ready = false;
+
+  // ValueNotifier for status text.
+  final _statusTextNotifier = ValueNotifier('Status: Not Connected');
+
+  // Dynamic entity settings state
+  bool _showsTrackLine = true;
+  bool _showsPreviousObservations = true;
```
jenmerritt:
```suggestion
  var _showsTrackLine = true;
  var _showsPreviousObservations = true;
```

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+
+  // Maximum observations.
+  double _maximumObservations = 5;
```
jenmerritt:
```suggestion
  var _maximumObservations = 5;
```
I guess you may need to make this `5.0` if a `double` is required, as opposed to an `int`.

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+
+  // Connection Status track.
+  bool _isTogglingConnection = false;
```
jenmerritt:
```suggestion
  var _isTogglingConnection = false;
```

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+    _dismissDynamicEntityCallout();
+
+    // Identify the first geo-element at the tap location in the dynamic entity layer.
+    final identifyResult = await _mapViewController.identifyLayer(
+      layer,
+      screenPoint: screenPoint,
+      tolerance: _identifyTolerance,
+    );
+
+    if (!mounted) return;
```
jenmerritt, line 226: I'm curious why this check is needed here specifically?

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+    final geoElements = identifyResult.geoElements;
+
+    // If identify failed or hit nothing, keep the callout dismissed.
+    if (identifyResult.error != null || geoElements.isEmpty) {
+      return;
+    }
+
+    final geoElement = geoElements.first;
+
+    // Prefer showing a callout for the DynamicEntity so it follows the moving vehicle.
```
jenmerritt:
```suggestion
    // Show a callout for the DynamicEntity so it follows the moving vehicle.
```
Wasn't sure about "prefer"? It looks like we either show a callout for the dynamic entity, or we don't.

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+    setState(() => _isTogglingConnection = true);
+
+    try {
+      // Decide action based on the current status at tap time.
+      if (service.connectionStatus == ConnectionStatus.connected) {
+        await service.disconnect();
+      } else {
+        await service.connect();
+      }
+    } catch (_) {
```
jenmerritt: There is an IDE warning on this line

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+    // Arcade expressions use fields from the SandyVehicles StreamServer schema:
+    // vehicle name, speed, heading, vehicle type, agency, point_x, point_y
+    final shown = _mapViewController.callout.showCalloutForGeoElement(
+      entity,
+      animated: false,
+      title: 'Vehicle',
+      detail: '',
+      titleExpression: r'$feature.vehiclename',
+      detailExpression:
+          r'concatenate('
```
jenmerritt: There is an IDE warning on this line

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+          r'"Speed: ", $feature.speed, " mph", '
+          r'"  Heading: ", $feature.heading, "°", '
+          r'"  Type: ", $feature.vehicletype, '
+          r'"  Agency: ", $feature.agency, '
+          r'"  (", Round($feature.point_x, 6), ", ", Round($feature.point_y, 6), ")"'
```
jenmerritt: There is an IDE warning on this line

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+          r')',
```
jenmerritt: There is an IDE warning on this line

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+                    width: double.infinity,
+                    child: OutlinedButton.icon(
+                      icon: const Icon(Icons.delete_outline),
+                      label: const Text('Purge All Observations'),
+                      onPressed: () async {
+                        final service = _streamService;
+                        if (service != null) {
+                          try {
+                            await service.purgeAll();
+                          } catch (_) {
```
jenmerritt: There is an IDE warning on this line

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+                            ScaffoldMessenger.of(context).showSnackBar(
```
jenmerritt: There is an IDE warning on this line

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+                              const SnackBar(
+                                content: Text('All observations purged.'),
+                              ),
+                            );
+                          }
+                        }
+
+                        if (mounted) Navigator.of(context).pop();
```
jenmerritt: There is an IDE warning on this line

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.png
jenmerritt, line 1: Could the content be a little tidier so like in the screengrab here, the "Type" is split across 2 lines. I'm fine with their being multiple key/value pairs on a line, but I think it could be a little tidier / more realistic to how it might be presented in an app.

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+                ),
+              ),
+          ],
+        ),
+      ),
+    );
+  }
+
+  Future<void> onMapViewReady() async {
+    // Create a map with a topographic basemap style.
```
jenmerritt:
```suggestion
    // Create a map with the streets basemap style.
```

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+            ),
+            // Display a progress indicator and prevent interaction until state is ready.
+            LoadingIndicator(visible: !_ready),
+            // Display a banner with the current status at the top.
+            if (_streamService != null)
+              SafeArea(
+                child: IgnorePointer(
+                  child: Container(
+                    padding: const EdgeInsets.all(10),
+                    color: Colors.white.withValues(alpha: 0.7),
```
jenmerritt: I think the opacity should be increased, I find it quite hard to read over the basemap labels. Or potentially consider a dark background instead of white.

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+          r')',
+    );
+
+    if (!shown) {
+      _mapViewController.callout.dismiss(animated: false);
+    }
+  }
+
+  // Dynamic Entity settings modal sheet.
+  void _showDynamicEntitySettings() {
```
jenmerritt, line 308: We have an 'X' button in the settings in other samples, so should look to include that here for consistency.

jenmerritt (commented):
Tested the sample - worked well.

Code comments:

lib/samples/add_dynamic_entity_layer/README.md
```diff
@@ -0,0 +1,47 @@
...
+Use `ArcGISStreamService` to manage the connection to the stream service and purge options to manage how much data is stored and maintained by the application. The dynamic entity layer will display the latest received observation, and you can set track display properties to determine how to display historical information for each dynamic entity. This includes the number of previous observations to show, whether to display track lines in-between previous observations, and setting renderers.
+
+## How to use the sample
+
+Use the controls to connect to or disconnect from the stream service, modify display properties in the dynamic entity layer, and purge all observations from the application.
+
+## How it works
+
+1. Create an `ArcGIStreamService` using a `Uri`.
+2. Set a `DynamicEntityFilter` on the stream service to limit the amount of data coming from the server.
```
PaulAllanSturm:
```suggestion
2. Set an `ArcGISStreamServiceFilter` on the stream service to limit the amount of data coming from the server.
```

lib/samples/add_dynamic_entity_layer/README.md
PaulAllanSturm: (and then move it up so the list is alphabetical)

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+class _AddDynamicEntityLayerState extends State<AddDynamicEntityLayer>
+    with SampleStateSupport {
+  // Create a controller for the map view.
+  final _mapViewController = ArcGISMapView.createController();
+
+  // A flag for when the map view is ready and controls can be used.
+  var _ready = false;
+
+  // ValueNotifier for status text.
+  final _statusTextNotifier = ValueNotifier('Status: Not Connected');
```
PaulAllanSturm: I don't think this is used anywhere in the UI, so it can be deleted

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
PaulAllanSturm: `TrackDisplayProperties` wants it to be an `int`, while `Slider` wants it to be a `double`. It makes more sense as an int (you can't have a fraction of an observation), so we should handle it as an int everywhere except when we're passing it in to `Slider`.

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
PaulAllanSturm: Yeah like in `find_route_in_mobile_map_package.dart` we use `Colors.black` background and `textTheme.customWhiteStyle` for the text.

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+                      mainAxisAlignment: MainAxisAlignment.center,
+                      children: [
+                        Expanded(
+                          child: StreamBuilder<ConnectionStatus>(
+                            stream: _streamService!.onConnectionStatusChanged,
+                            initialData: _streamService!.connectionStatus,
+                            builder: (context, snapshot) {
+                              final status =
+                                  snapshot.data ??
+                                  _streamService!.connectionStatus;
```
PaulAllanSturm: I think the proper way to use StreamBuilder's `builder` callback is to only rely on the information passed in via the `snapshot` parameter. We shouldn't access member variables. So instead of `_streamService!.connectionStatus` we should simply use `.disconnected` here.

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+    _streamService = service;
+    _dynamicEntityLayer = layer;
+
+    // Initialize UI state from layer defaults.
+    // This is so that settings reflects actual layer state.
+    _showsTrackLine = layer.trackDisplayProperties.showTrackLine;
+    _showsPreviousObservations =
+        layer.trackDisplayProperties.showPreviousObservations;
+    _maximumObservations = layer.trackDisplayProperties.maximumObservations
+        .toDouble();
```
PaulAllanSturm: These three lines should be inside a `setState`.

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
PaulAllanSturm: The general rule is: after performing async work in a callback, you best check `mounted` before doing anything afterwards that updates the UI. The scenario we're avoiding is if the user taps on the map and then quickly backs out of the sample before the "identifyLayer" operation completes. In that scenario the widget has been disposed, so we need to return early from this method.

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+      }
+    } catch (_) {
+      // Snackbar to display the connection status.
+      if (mounted) {
+        ScaffoldMessenger.of(context).showSnackBar(
+          const SnackBar(content: Text('Failed to change connection status.')),
+        );
+      }
+    } finally {
+      if (mounted) {
```
PaulAllanSturm: This class uses `with SampleStateSupport`, which cleverly overrides `setState()` with a built-in `mounted` check. So we don't actually need to guard this particular line with the `if (mounted)`.

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
PaulAllanSturm: you can use a single multi-line string with `'''`, rather than piecing together a bunch of single-line strings. So it would look something like
```dart
detailExpression: r'''
concatenate(
"Speed: ", $feature.speed, "mph",
"  Heading:", $feature.heading, "°",
etc...
)
'''
```

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+          r'"Speed: ", $feature.speed, " mph", '
+          r'"  Heading: ", $feature.heading, "°", '
+          r'"  Type: ", $feature.vehicletype, '
+          r'"  Agency: ", $feature.agency, '
+          r'"  (", Round($feature.point_x, 6), ", ", Round($feature.point_y, 6), ")"'
+          r')',
+    );
+
+    if (!shown) {
+      _mapViewController.callout.dismiss(animated: false);
```
PaulAllanSturm: you could use `_dismissDynamicEntityCallout()` since you have it

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
PaulAllanSturm: Also I think we'd prefer a non-modal sheet, rather than a modal sheet. That way you can navigate around the map while the sheet is up. We have a `BottomSheetSettings` helper widget that gives you the close button and some other things. So this overall would be better using the `bottomSheet:` parameter of `Scaffold` and our `BottomSheetSettings` widget. One example is `add_feature_layer_with_time_offset.dart`.

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,454 @@
...
+          divisions: divisions,
+          label: value.toStringAsFixed(0),
+          onChanged: onChanged,
+        ),
+      ],
+    );
+  }
+}
+
+extension ConnectionStatusUi on ConnectionStatus {
```
PaulAllanSturm: We can just use an unnamed extension when we're adding stuff purely for our own internal use.
```dart
extension on ConnectionStatus {
```
That saves us the trouble of having to come up with a meaningless meaningful name for the extension.

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
harish-kunchala: Basically we have to specify the type of exception that occurs. But since there are multiple type of exceptions that can occur, either due to our `ConnectionStatus` or default network exceptions like `SocketException`, `HttpException` or `ClientException`, I basically just specified a dynamic object type to cover all of them. Our IDE is asking to define the exact type of exception we want to catch.

PaulAllanSturm: In that case I think you just catch the base `Exception` type, which covers them all.

PaulAllanSturm: yeah apply_map_algebra.dart has an example
```dart
    } on Exception catch (_) {
```

lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart
```diff
@@ -0,0 +1,441 @@
...
+    }
+  }
+
+  // Show the vehicle dynamic entity callout.
+  void _showVehicleCallout(DynamicEntity entity) {
+    // Arcade expressions use fields from the SandyVehicles StreamServer schema:
+    // vehicle name, speed, heading, vehicle type, agency, point_x, point_y
+    final shown = _mapViewController.callout.showCalloutForGeoElement(
+      entity,
+      animated: false,
```
martinoyovo, line 287: `animated` is false by default, we can remove it.

harish-kunchala: Oh no. It seems be set to true.

```callout.dart
 bool showCalloutForGeoElement(
    GeoElement geoElement, {
    String? title,
    String? detail,
    ArcGISPoint? tapLocation,
    bool animated = true,
    ```

Comment (martinoyovo) [reply lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart:287]: oh wow, you're right.

File: lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart (hunk 11 above)
Comment (jenmerritt) [reply lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart:226]: I understand what it means, my question was more why here specifically. In previous usages I've seen it / have expected it to be called before a UI update, as opposed to right after an async function. The UI isn't updated until line 249.

**Review by jenmerritt — APPROVED** [review]

File: lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart (hunk 11 above)
Comment (PaulAllanSturm) [reply lib/samples/add_dynamic_entity_layer/add_dynamic_entity_layer.dart:226]: Oh, right right right. I suppose there are two options:
a) check `mounted` immediately after an `async` task to clear the condition
b) guard all downstream UI access with a `mounted` check

I think "a" makes more sense as a general rule. The cause-and-effect is tightly bound together and clear. With "b", you have to trace down through potentially multiple branching logics and method jumps to catch the condition at every possible end point, which would breed this reflexive/defensive habit of sprinkling `mounted` checks all over and not knowing when/if/why they're relevant.

File: lib/common/bottom_sheet_settings.dart [hunk 31]
```diff
@@ -40,7 +40,7 @@ class BottomSheetSettings extends StatelessWidget {
     return Container(
       padding: EdgeInsets.fromLTRB(
         20,
-        20,
+        5,
```
PaulAllanSturm: I don't think we should make this change as part of this sample -- it might have downstream implications (like automation screen captures) and maybe we might want to address this a little differently. I do agree that there's too much padding, but we should consider it independently of this particular sample.

harish-kunchala: Agreed. I'll raise a separate issue to modify bottom sheet padding.

PR #460, opened 2026-04-01
Updated Bottom sheet top padding to 5.

harish-kunchala:
- Updated Bottom Sheet top padding from 20 to 5.

padding-20
padding-5


harish-kunchala:
Moved to next release.

PR #462, opened 2026-04-02
Update share_plus for iOS 26 support

PaulAllanSturm:
`share_plus` has a fix in its version 12 for a problem on iOS 26. It also deprecates `Share.share()` and instructs us to move to `SharePlus.instance.share()`.

Also I noticed we were using "localHistory.geodatabase" as the filename, but the design intended "LocationHistory.geodatabase".


harish-kunchala:
I noticed that after the Share dialog pops up, if I dismiss it. It clears out the mobile Geodatabase. Is it the intended behavior ?

harish-kunchala (approved):

martinoyovo (approved):

PaulAllanSturm:
> I noticed that after the Share dialog pops up, if I dismiss it. It clears out the mobile Geodatabase. Is it the intended behavior ?

We kind of assume you successfully shared the file somewhere, and then we intentionally start over so you can go through the process again. The design isn't really specific on the point, and we don't want to add a lot of complexity around the share sheet, because that's not the focus of the sample.

PR #476, opened 2026-04-13
Updated Color theme to reflect light and dark mode dynamically

harish-kunchala:
- Added dark theme values into `theme_data.dart`.
- Added `darkTheme:` property into our main and sample_runner.

# Output:

## Samples Overview:
Samples Overview Light Mode
Samples Overview Dark Mode

## Samples List
Samples List Light Mode

Samples List Dark Mode

## Inside a Sample
Bottom Sheet Settings Light Mode

Bottom Sheet Settings Dark Mode

### DIFF
--- a/lib/common/theme_data.dart
+++ b/lib/common/theme_data.dart
@@ -25,75 +25,118 @@ final darkColorScheme = ColorScheme.fromSeed(
   brightness: Brightness.dark,
 );
 
-// Light Theme
-final sampleViewerLightTheme = ThemeData(
-  brightness: Brightness.light,
-  colorScheme: lightColorScheme,
-  primaryColor: lightColorScheme.primary,
+// This shared builder ensures light and dark themes stay in sync,
+// differing only by their ColorScheme. Any component theme changes
+// automatically apply to both modes.
+ThemeData _buildTheme(ColorScheme colorScheme) {
+  return ThemeData(
+    brightness: colorScheme.brightness,
+    colorScheme: colorScheme,
 
-  // Application bar theme
-  appBarTheme: AppBarTheme(backgroundColor: lightColorScheme.inversePrimary),
-
-  // Text theme
-  textTheme: const TextTheme(labelMedium: TextStyle(color: Colors.deepPurple)),
-
-  // Button themes
-  elevatedButtonTheme: ElevatedButtonThemeData(
-    style: ElevatedButton.styleFrom(
-      disabledBackgroundColor: Colors.white.withValues(alpha: 0.6),
+    // Application bar theme.
+    appBarTheme: AppBarTheme(
+      backgroundColor: colorScheme.inversePrimary,
+      foregroundColor: colorScheme.onSurface,
+      elevation: 0,
+      centerTitle: false,
     ),
-  ),
 
-  floatingActionButtonTheme: FloatingActionButtonThemeData(
-    backgroundColor: lightColorScheme.primaryContainer,
-  ),
-
-  dropdownMenuTheme: DropdownMenuThemeData(
-    inputDecorationTheme: const InputDecorationTheme(
-      outlineBorder: BorderSide(width: 0),
-      border: InputBorder.none,
+    // Button themes
+    elevatedButtonTheme: ElevatedButtonThemeData(
+      style: ElevatedButton.styleFrom(
+        backgroundColor: colorScheme.primary,
+        foregroundColor: colorScheme.onPrimary,
+        // Use ColorScheme for disabled states (works in light & dark)
+        disabledBackgroundColor: colorScheme.onSurface.withValues(alpha: 0.12),
+        disabledForegroundColor: colorScheme.onSurface.withValues(alpha: 0.38),
+        elevation: 2,
+        shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(8)),
+      ),
     ),
-    menuStyle: MenuStyle(elevation: WidgetStateProperty.all(6)),
-  ),
 
-  // Icon theme
-  iconTheme: IconThemeData(color: lightColorScheme.primary),
-);
-
-// Dark Theme
-final sampleViewerDarkTheme = ThemeData(
-  brightness: Brightness.dark,
-  colorScheme: darkColorScheme,
-  primaryColor: darkColorScheme.primary,
-
-  // Application bar theme
-  appBarTheme: AppBarTheme(backgroundColor: darkColorScheme.inversePrimary),
-
-  // Text theme - adjusted for dark mode
-  textTheme: TextTheme(labelMedium: TextStyle(color: darkColorScheme.primary)),
-
-  // Button themes
-  elevatedButtonTheme: ElevatedButtonThemeData(
-    style: ElevatedButton.styleFrom(
-      disabledBackgroundColor: Colors.black.withValues(alpha: 0.3),
+    textButtonTheme: TextButtonThemeData(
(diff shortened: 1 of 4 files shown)


lib/common/theme_data.dart
```diff
@@ -16,30 +16,86 @@
...
 
-final sampleViewerTheme = ThemeData(
-  // color scheme
-  primaryColor: colorScheme.primary,
-  primaryColorLight: Colors.deepPurple[200],
-  disabledColor: Colors.grey,
-  colorScheme: colorScheme,
+// Light Theme
+final sampleViewerLightTheme = ThemeData(
+  useMaterial3: true,
```
PaulAllanSturm: We don't need to specify `useMaterial3: true` any more. That's the default, and this flag will eventually be deprecated. So we should remove `useMaterial3: true`.

lib/common/theme_data.dart
```diff
@@ -16,30 +16,86 @@
...
+    menuStyle: MenuStyle(elevation: WidgetStateProperty.all(6)),
   ),
+
+  // Icon theme
+  iconTheme: IconThemeData(color: lightColorScheme.primary),
+);
+
+// Dark Theme
+final sampleViewerDarkTheme = ThemeData(
+  useMaterial3: true,
```
PaulAllanSturm: Likewise we can remove `useMaterial3: true` here too.

PaulAllanSturm (approved):

lib/common/theme_data.dart
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright 2024 Esri
+// Copyright 2026 Esri
```
jenmerritt: No need to revert this as it's not a big deal, but just to say for future ref that these licenses are valid from the date the file is created, so we don't need to actively update/maintain the license dates.

jenmerritt (commented):
The screengrabs LGTM. Keen for @martinoyovo to take a look at the implementation as well for his Flutter experience.

For verification I don't think there'll be any impact on automated samples tests looking at the PR, but when it's merged worth keeping in mind to make sure Kathleen is aware just incase.

lib/common/theme_data.dart
```diff
@@ -16,30 +16,73 @@
...
+final lightColorScheme = ColorScheme.fromSeed(seedColor: Colors.deepPurple);
 
-final sampleViewerTheme = ThemeData(
-  // color scheme
-  primaryColor: colorScheme.primary,
-  primaryColorLight: Colors.deepPurple[200],
-  disabledColor: Colors.grey,
-  colorScheme: colorScheme,
+// Dark Theme Color Scheme
+final darkColorScheme = ColorScheme.fromSeed(
```
martinoyovo, line 23: I think we should collapse the duplicated light and dark theme construction into one shared builder. Right now both themes differ mostly by scheme, so a single `buildTheme` helper would reduce drift and make future changes safer

harish-kunchala: Yeah, but since Material provides `darkTheme` variable. I figured it would be simpler to separate both light and dark color schemes.

lib/common/theme_data.dart
```diff
@@ -16,30 +16,73 @@
...
+  primaryColor: lightColorScheme.primary,
 
-  // application bar theme
-  appBarTheme: AppBarTheme(backgroundColor: colorScheme.inversePrimary),
+  // Application bar theme
+  appBarTheme: AppBarTheme(backgroundColor: lightColorScheme.inversePrimary),
 
-  // text theme
+  // Text theme
   textTheme: const TextTheme(labelMedium: TextStyle(color: Colors.deepPurple)),
```
martinoyovo: Setting `labelMedium` and all icons to primary can create poor contrast on cards, dialogs, sheets, and other surfaces where primary is not the right foreground color.

lib/common/theme_data.dart
```diff
@@ -16,30 +16,73 @@
...
+// Light Theme
+final sampleViewerLightTheme = ThemeData(
+  brightness: Brightness.light,
+  colorScheme: lightColorScheme,
+  primaryColor: lightColorScheme.primary,
 
-  // application bar theme
-  appBarTheme: AppBarTheme(backgroundColor: colorScheme.inversePrimary),
+  // Application bar theme
+  appBarTheme: AppBarTheme(backgroundColor: lightColorScheme.inversePrimary),
```
martinoyovo: We should reconsider the app bar color role. inversePrimary is not the most typical background choice for AppBar. surface/onSurface or primaryContainer/onPrimaryContainer are usually more predictable.

lib/common/theme_data.dart
```diff
@@ -16,30 +16,73 @@
...
+);
+
+// Dark Theme
+final sampleViewerDarkTheme = ThemeData(
+  brightness: Brightness.dark,
+  colorScheme: darkColorScheme,
+  primaryColor: darkColorScheme.primary,
+
+  // Application bar theme
+  appBarTheme: AppBarTheme(backgroundColor: darkColorScheme.inversePrimary),
```
martinoyovo: Same comment on the app bar color here.

lib/common/theme_data.dart
```diff
@@ -16,30 +16,73 @@
...
+  // Application bar theme
+  appBarTheme: AppBarTheme(backgroundColor: lightColorScheme.inversePrimary),
 
-  // text theme
+  // Text theme
   textTheme: const TextTheme(labelMedium: TextStyle(color: Colors.deepPurple)),
 
-  // button theme
+  // Button themes
   elevatedButtonTheme: ElevatedButtonThemeData(
```
martinoyovo: I think here we should make disabled button colors theme-derived instead of manually white/black with alpha in. I suggest this:  `colorScheme.onSurface.withValues(alpha: 0.12);`

lib/common/theme_data.dart
martinoyovo: I mean, the only thing that's changing from both themes is the `colorScheme`. We could define a helper `_buildSampleViewerTheme(ColorScheme colorScheme)` and that'd just return the `ThemeData` based on the colorScheme.

lib/common/theme_data.dart
```diff
@@ -16,41 +16,127 @@
...
+ThemeData _buildTheme(ColorScheme colorScheme) {
+  return ThemeData(
+    brightness: colorScheme.brightness,
+    colorScheme: colorScheme,
 
-  // application bar theme
-  appBarTheme: AppBarTheme(backgroundColor: colorScheme.inversePrimary),
+    // Application bar theme.
+    appBarTheme: AppBarTheme(
+      backgroundColor: colorScheme.inversePrimary,
```
martinoyovo: We can use: `backgroundColor: colorScheme.surface`

lib/common/theme_data.dart
```diff
@@ -16,41 +16,127 @@
...
+// Dark Theme Color Scheme
+final darkColorScheme = ColorScheme.fromSeed(
+  seedColor: Colors.deepPurple,
+  brightness: Brightness.dark,
+);
+
+// This shared builder ensures light and dark themes stay in sync,
+// differing only by their ColorScheme. Any component theme changes
+// automatically apply to both modes.
+ThemeData _buildTheme(ColorScheme colorScheme) {
```
martinoyovo, line 29: I think we could just limit the theme data to what was previously there:

- appBarTheme
- elevatedButtonTheme
- floatingActionButtonTheme
- dropdownMenuTheme

martinoyovo (approved):

PR #488, opened 2026-04-29
Harish/dark theme UI fixes

harish-kunchala:
- Added DarkMode for Callout
- Added Dynamic Theme Support for DropDown Menu Items.
- Softer DarkMode colors.

### DIFF
--- a/lib/common/callout_theme.dart
+++ b/lib/common/callout_theme.dart
@@ -62,6 +62,8 @@ extension ThemedCalloutStyle on CalloutStyle {
           contentPadding ??
           const EdgeInsets.symmetric(horizontal: 16, vertical: 12),
       maxWidth: maxWidth ?? 300,
+      //CalloutStyle default Offset value.
+      offset: offset ?? const Offset(0, -8),
       minWidth: 100,
     );
   }
--- a/lib/common/callout_theme.dart
+++ b/lib/common/callout_theme.dart
@@ -1,3 +1,18 @@
+// Copyright 2026 Esri
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
 import 'package:arcgis_maps/arcgis_maps.dart';
 import 'package:flutter/material.dart';
 
--- a/lib/common/callout_theme.dart
+++ b/lib/common/callout_theme.dart
@@ -59,12 +59,10 @@ extension ThemedCalloutStyle on CalloutStyle {
 
       // Layout properties.
       contentPadding:
-      contentPadding ??
+          contentPadding ??
           const EdgeInsets.symmetric(horizontal: 16, vertical: 12),
-      offset: offset ?? const Offset(0, -35),
       maxWidth: maxWidth ?? 300,
       minWidth: 100,
     );
   }
-
 }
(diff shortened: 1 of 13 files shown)


lib/common/callout_theme.dart
PaulAllanSturm, line 1: This file needs our usual copyright header (the Apache License one).

lib/widgets/download_progress_card.dart
PaulAllanSturm, line 1: This file also needs the copyright header.

lib/common/callout_theme.dart
```diff
@@ -0,0 +1,55 @@
...
+      ),
+      detailTextStyle: textTheme.bodyMedium?.copyWith(
+        color: colorScheme.onSurfaceVariant,
+      ),
+
+      // Layout properties (use provided values or sensible defaults)
+      contentPadding:
+          contentPadding ??
+          const EdgeInsets.symmetric(horizontal: 16, vertical: 12),
+      offset: offset ?? const Offset(0, -35),
```
PaulAllanSturm: I don't think we want to use `-35` here. I think that was a special case for "Find address with reverse geocode" only.

CalloutStyle already provides a sensible default, so this line could just be:
```dart
offset: offset
```

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode.dart
```diff
@@ -133,10 +135,7 @@ class _FindAddressWithReverseGeocodeState
     _mapViewController.callout.showAt(
       normalizedTapPoint,
       detail: combinedString,
-      style: CalloutStyle(
-        contentPadding: const EdgeInsets.symmetric(horizontal: 16, vertical: 8),
-        offset: const Offset(0, -35),
-      ),
+      style: ThemedCalloutStyle.themed(context),
```
PaulAllanSturm: Carrying on from the previous comment, I think we still need to pass in `offset: const Offset(0, -35)` here. This is the only sample that should have the special `-35` y offset. All other samples should use the default, which should be the `-8` y offset.

lib/widgets/download_progress_card.dart
harish-kunchala: Done.

lib/common/callout_theme.dart
harish-kunchala: Done.

lib/common/callout_theme.dart
harish-kunchala: Done.

lib/samples/find_address_with_reverse_geocode/find_address_with_reverse_geocode.dart
harish-kunchala: Done.

lib/widgets/download_progress_card.dart
```diff
@@ -0,0 +1,115 @@
+// Copyright 2024 Esri
```
PaulAllanSturm: 2024 should be 2026

lib/common/callout_theme.dart
PaulAllanSturm: I'm still seeing the `-35` here. I was expecting just:
```dart
offset: offset,
```
We want it to fall through to the CalloutStyle default value, which is `Offset(0, -8)`. We don't want `-35` as the fallback/default. The `-35` was just for that one special case.

lib/common/callout_theme.dart
```diff
@@ -0,0 +1,68 @@
...
+        fontWeight: FontWeight.w600,
+      ),
+      detailTextStyle: textTheme.bodyMedium?.copyWith(
+        color: colorScheme.onSurfaceVariant,
+      ),
+
+      // Layout properties.
+      contentPadding:
+          contentPadding ??
+          const EdgeInsets.symmetric(horizontal: 16, vertical: 12),
```
PaulAllanSturm, line 63: but now the `offset` parameter is ignored entirely. It doesn't get passed in to the `CalloutStyle` at all. It should be passed in to the `CalloutStyle` constructor here.
```dart
 offset: offset,
```

harish-kunchala: You are right. I've set a the default offSet value here since it can be nullable.

PaulAllanSturm (approved):

martinoyovo (approved):

PR #493, opened 2026-05-06
Add license placeholder

jenmerritt:
Makes it possible for us to set a license string on daily builds.

Makes it clear this is optional to developers to run the sample viewer.

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -31,6 +31,12 @@ void main() async {
   // const apiKey = ''; // Your API Key here.
   ArcGISEnvironment.apiKey = apiKey;
 
+  // (Optional) Supply a license key using the --dart-define-from-file command line argument.
+  const licenseKey = String.fromEnvironment('LICENSE_KEY');
+  if (licenseKey.isNotEmpty) {
+    ArcGISEnvironment.setLicenseUsingKey(licenseKey);
+  }
+
   WidgetsFlutterBinding.ensureInitialized();
 
   final prefs = await SharedPreferences.getInstance();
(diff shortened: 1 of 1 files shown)


PaulAllanSturm (approved):

martinoyovo (approved):

jenmerritt:
This might need an additional update to handle extensions, awaiting confirmation. So not merging yet.

jenmerritt:
Updated to include extensions

martinoyovo (approved):

lib/main.dart
```diff
@@ -31,6 +31,21 @@ void main() async {
...
 
+  // (Optional) Supply a license key using the --dart-define-from-file command line argument.
+  const licenseKey = String.fromEnvironment('LICENSE_KEY');
+  const advancedEditingExtension = String.fromEnvironment(
+    'ADVANCED_EDITING_EXTENSION',
+  );
+  const analysisExtension = String.fromEnvironment('ANALYSIS_EXTENSION');
+  if (licenseKey.isNotEmpty &&
+      advancedEditingExtension.isNotEmpty &&
+      analysisExtension.isNotEmpty) {
```
PaulAllanSturm, line 42: This requires all 3 environment variables to be set before any of them are used. Do we want to support the use case where someone provides a license key with no (or only 1) of the extensions?

jenmerritt: I haven't allowed for that, I was just doing it for the use case I have at the moment which is for the daily builds.

PaulAllanSturm (approved):

PR #498, opened 2026-05-12
Cs/SearchDictionarySymbols Sample

changanxian:


lib/samples/search_symbol_style_dictionary/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+To search for certain symbols, enter text into one or more search boxes and click 'Search for symbols'. Results are shown in a list. Pressing 'Clear' will reset the search.
+
+## How it works
+
+1. Create a `DictionarySymbolStyle` from the military symbology `.stylx` file using `DictionarySymbolStyle.withFileUri(...)`.
+2. Create `SymbolStyleSearchParameters`.
+3. Add values to the `names`, `tags`, `symbolClasses`, `categories`, and `keys` lists on the search parameters.
+4. Search for matching symbols with `DictionarySymbolStyle.searchSymbols(...)`.
+5. Get the `Symbol` from each returned `SymbolStyleSearchResult`.
```
PaulAllanSturm: We had to map the name "Symbol" to "ArcGISSymbol" so this needs to be renamed here, in the Relevant API list below, and in the README.metadata.json. And moved into alphabetical order.

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.png
PaulAllanSturm, line 1: This might be too cramped vertically, when the keyboard is up and especially in landscape orientation. If you have the "more" panel revealed and have the keyboard up, you get overflow errors. In landscape orientation, you get overflow errors and most of the controls are inaccessible.

If we wanted to make this work in all combinations, I think we'd have to have two pages, like in the Swift version. The first page would be the search interface, and it would be in a scrollable frame. The second page would be just the search results (and a back button).

PaulAllanSturm (commented):
some initial comments ...

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
```diff
@@ -0,0 +1,473 @@
...
+
+  @override
+  State<SearchSymbolStyleDictionary> createState() =>
+      _SearchSymbolStyleDictionaryState();
+}
+
+class _SearchSymbolStyleDictionaryState
+    extends State<SearchSymbolStyleDictionary>
+    with SampleStateSupport {
+  final _nameController = TextEditingController();
```
PaulAllanSturm, line 35: Descriptive comment above this group of member variables, maybe something like
```dart
  // Text editing controllers for the search fields.
```

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
```diff
@@ -0,0 +1,473 @@
...
+  final _tagController = TextEditingController();
+  final _symbolClassController = TextEditingController();
+  final _categoryController = TextEditingController();
+  final _keyController = TextEditingController();
+
+  var _ready = true;
```
PaulAllanSturm, line 42: Maybe a comment above `_ready` like
```dart
  // A flag to prevent interaction when a search is in progress.
```

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
```diff
@@ -0,0 +1,473 @@
...
+
+      // Download the sample data.
+      final listPaths = GoRouter.of(context).state.extra! as List<String>;
+      final file = File(listPaths.first);
+      if (!file.existsSync() || file.lengthSync() == 0) {
+        _statusMessage =
+            'Symbol style dictionary file not found or empty at path: ${file.path}';
+        setState(() => _ready = true);
+        return;
+      }
```
PaulAllanSturm: Normally we don't do any defensive error checking on `GoRouter.of(context).state.extra`. We just assume that the download manager did what it's supposed to. Our approach maybe isn't very robust, but we want this code to act as a tutorial for our API and not get distracted by other concerns.

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
```diff
@@ -0,0 +1,473 @@
...
+                        style: Theme.of(context).textTheme.bodyMedium,
+                      ),
+                      _SearchFields(
+                        nameController: _nameController,
+                        tagController: _tagController,
+                        symbolClassController: _symbolClassController,
+                        categoryController: _categoryController,
+                        keyController: _keyController,
+                      ),
+                      Wrap(
```
martinoyovo: This keeps the Search and Clear buttons active while `_ready` is false. Multiple taps can start overlapping searches, and whichever finishes last wins, possibly showing stale results. 

I think we should disable controls while loading.

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
```diff
@@ -0,0 +1,473 @@
...
+      final symbol = await result.getSymbol();
+      final arcgisImage = await _getSwatchImage(symbol);
+      listSymbols.add(
+        _SymbolSearchPreview(
+          name: result.name,
+          tags: result.tags,
+          symbolClass: result.symbolClass,
+          category: result.category,
+          key: result.key,
+          swatch: Image.memory(arcgisImage.getEncodedBuffer()),
```
PaulAllanSturm: We have a utility `class SwatchImage` in `lib/common/swatch_image.dart`. Could we use that here? We would just pass in the `symbol` rather than using `_getSwatchImage()`. `SwatchImage` loads the swatch image asynchronously, which would allow the list to load quicker.

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
```diff
@@ -0,0 +1,473 @@
...
+              ),
+            ),
+            LoadingIndicator(visible: !_ready),
+          ],
+        ),
+      ),
+    );
+  }
+
+  Future<void> _performSearch() async {
```
martinoyovo: Could we try/catch this? That'd allow `_ready` to be restored and `_statusMessage` to be updated on failure.

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
PaulAllanSturm: I think the `LoadingIndicator` below prevents any interaction with the buttons. It covers the whole screen and is in a `Stack` above everything else.

martinoyovo: Makes sense, thank you

lib/samples/search_symbol_style_dictionary/README.md
```diff
@@ -12,20 +14,21 @@ You can use support for military symbology to allow users to report changes in t
 
 By default, leaving the fields blank and hitting search will find all symbols.
 
-To search for certain symbols, enter text into one or more search boxes and click 'Search for symbols'. Results are shown in a list. Pressing 'Clear' will reset the search.
+To search for certain symbols, enter text into one or more search boxes and click 'Search for symbols'. Results are shown on a separate page, and the back button returns to the search form. Pressing 'Clear' will reset the search.
```
PaulAllanSturm: "click" should be "tap"

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
```diff
@@ -0,0 +1,560 @@
...
+    addIfNotEmpty(searchFilter.categories, _categoryController.text);
+    addIfNotEmpty(searchFilter.keys, _keyController.text);
+
+    return searchFilter;
+  }
+
+  Future<void> _performSearch() async {
+    FocusManager.instance.primaryFocus?.unfocus();
+
+    // prevent multiple simultaneous searches
```
PaulAllanSturm: The comments in this method need to be capitalized with a period at the end.

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
```diff
@@ -0,0 +1,560 @@
...
+    final searchFilter = SymbolStyleSearchParameters();
+    // A helper function to add a search parameter to the search filter if the parameter value is not empty.
+    void addIfNotEmpty(List<String> target, String value) {
+      final trimmedValue = value.trim();
+      if (trimmedValue.isNotEmpty) {
+        target.add(trimmedValue);
+      }
+    }
+
+    // add search parameters from the search fields if they are not empty
```
PaulAllanSturm: This comment should be capitalized with a period at the end.

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
```diff
@@ -0,0 +1,560 @@
...
+    );
+  }
+
+  // Load the symbol style dictionary from the file URI passed in the route parameters.
+  Future<void> _loadDictionarySymbolStyle() async {
+    // Download the sample data.
+    try {
+      final listPaths = GoRouter.of(context).state.extra! as List<String>;
+      final file = File(listPaths.first);
+      // load the symbol style dictionary from the file URI
```
PaulAllanSturm: This comment should be capitalized with a period at the end. (I wonder if we can have an AI skill that validates our comment style?)

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
```diff
@@ -0,0 +1,560 @@
...
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      body: SafeArea(
+        child: Stack(
+          children: [
+            if (_showResultsPage)
+              _SearchResultsPage(
```
PaulAllanSturm, line 77: Maybe a comment here like,
```dart
  // When search results are available, show the Search Results Page.
```

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
```diff
@@ -0,0 +1,560 @@
...
+                resultCount: _resultCount,
+                results: _results,
+                statusMessage: _statusMessage,
+                onBack: _showSearchPage,
+              )
+            else
+              _SearchFormPage(
```
PaulAllanSturm, line 85:
```
  // Otherwise, show the Search Form Page.
```

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
```diff
@@ -0,0 +1,560 @@
...
+    });
+  }
+
+  void _showSearchPage() {
+    FocusManager.instance.primaryFocus?.unfocus();
+    setState(() => _showResultsPage = false);
+  }
+}
+
+// A widget has a list of search fields for symbol style dictionary search parameters.
```
PaulAllanSturm: Add "that" in between "... widget that has ..."

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
```diff
@@ -0,0 +1,560 @@
...
+    required this.key,
+    this.swatch,
+  });
+
+  final String name;
+  final List<String> tags;
+  final String symbolClass;
+  final String category;
+  final String key;
+  final SwatchImage? swatch;
```
PaulAllanSturm: this doesn't need to be nullable -- we always pass in a real SwatchImage

lib/samples/search_symbol_style_dictionary/search_symbol_style_dictionary.dart
changanxian: >I wonder if we can have an AI skill that validates our comment style?

It is a good idea.

PaulAllanSturm: I added a comment to issue 5268

PaulAllanSturm (approved):

martinoyovo (approved):

PR #500, opened 2026-05-13
Add UI to Sample Viewer to set an API key as a user

martinoyovo:
There’s intentionally no validation or verification of the API key here. The implementation assumes users provide a valid key, which is consistent with the existing Qt implementation. The only processing we apply is trimming leading/trailing whitespace to avoid issues caused by accidental spaces.

By default the API key is not persisted, the implementation is a session-only override.

### DIFF
--- a/lib/common/api_key_manager.dart
+++ b/lib/common/api_key_manager.dart
@@ -20,13 +20,19 @@ import 'package:flutter/foundation.dart';
 class ApiKeyManager {
   ApiKeyManager._();
 
-  // The API key provided at build time using `--dart-define`.
-  static const _buildTimeApiKey = String.fromEnvironment('API_KEY');
+  // The API key provided by the app at startup.
+  static var _buildTimeApiKey = '';
 
   // Whether the active API key is a session-only override.
   static final isUsingOverride = ValueNotifier(false);
 
-  // Applies the build-time API key and clears any session-only override state.
+  // Stores and applies the default API key for the app session.
+  static void setBuildTimeKey(String buildTimeKey) {
+    _buildTimeApiKey = buildTimeKey;
+    applyBuildTimeApiKey();
+  }
+
+  // Applies the stored build-time API key and clears any override state.
   static void applyBuildTimeApiKey() {
     ArcGISEnvironment.apiKey = _buildTimeApiKey;
     isUsingOverride.value = false;
--- a/lib/common/api_key_manager.dart
+++ b/lib/common/api_key_manager.dart
@@ -1,0 +1,46 @@
+//
+// Copyright 2026 Esri
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


lib/common/api_key_manager.dart
jenmerritt, line 1: > By default the API key is not persisted, the implementation is a session-only override.

Agreed - session-only is definitely sufficient.

lib/main.dart
```diff
@@ -25,11 +26,12 @@ import 'package:flutter_localizations/flutter_localizations.dart';
...
-  const apiKey = String.fromEnvironment('API_KEY');
-  // Alternatively, replace the above line with the following and hard-code your apiKey here:
-  // const apiKey = ''; // Your API Key here.
-  ArcGISEnvironment.apiKey = apiKey;
+  // Initialize Flutter before reading preferences and bundled sample metadata.
+  WidgetsFlutterBinding.ensureInitialized();
+
+  // Use the build-time API key by default. The About screen can set a
+  // session-only override later without revealing the active key.
+  ApiKeyManager.applyBuildTimeApiKey();
```
jenmerritt: So with this change we've lost the ability for a developer looking to run the samples to set their API key directly - this is laid out in the readme, so if we did remove that, we'd need to be updating our supporting documentation. However I don't think we should be removing this option:
```
 // Alternatively, replace the above line with the following and hard-code your apiKey here:
 // const apiKey = ''; // Your API Key here.
```

I'm wondering if this is a bit over-engineered, or maybe I'm being naive. Do we need an API key manager? Is it that complex? Can we not just have the UI call `ArcGISEnvironment.apiKey = theApiKeyTypedIn;`

PaulAllanSturm: We need a "currently active key", a "default/build-time key", and a flag that tracks what's going on (with a ValueNotifier to integrate with the UI). That's enough to bundle up into a nice little class.

Probably we just need to incorporate the "set the key in code" use case a little better. Perhaps we:
1. add a static `ApiKeyManager.setBuildTimeKey(String buildTimeKey)` instead of having ApiKeyManager look it up from the environment
2. In `main`, set it up like this:
```dart
  const apiKey = String.fromEnvironment('API_KEY');
  // Alternatively, replace the above line with the following and hard-code your apiKey here:
  // const apiKey = ''; // Your API Key here.
  ApiKeyManager.setBuildTimeKey(apiKey);
```

jenmerritt: Ok fair enough. I thought this would just literally be a box and button to override the set API key on the ArcGISEnvironment. I didn't think it needed to be tracked as I can't see why we'd need to get the current key just like we didn't previously. But maybe it does need to be tracked in the app now.

martinoyovo: Done, I've refined the API key manager build-time setup!

PaulAllanSturm (approved):

jenmerritt (approved):
Make sure to provide clear verification instructions - we'll want to verify the UI on both platforms, and also verify setting a valid and an invalid API key - and anything else that is worth verification.

PR #503, opened 2026-05-15
Analyze hotspots sample

martinoyovo:


lib/samples/analyze_hotspots/analyze_hotspots.dart
```diff
@@ -0,0 +1,288 @@
...
+      );
+
+    // Set the map on the map view controller.
+    _mapViewController.arcGISMap = map;
+
+    try {
+      // Load the geoprocessing task so the Analyze button can create jobs from it.
+      await _hotspotTask.load();
+      // Enable the sample controls after the geoprocessing task loads.
+      if (mounted) setState(() => _ready = true);
```
PaulAllanSturm: We did a clever thing with setState and showMessageDialog -- we overrode setState with our own setState that checks `mounted` before calling the real setState. That means we don't have to check `if (mounted)` when we call setState. The AI doesn't realize this and so it's just wrong when it complains about not checking `mounted`. So we can remove `if (mounted)` from this line, as well as in the Exception block below, and in the `finally` block of `analyzeHotspots()` further down.

lib/samples/analyze_hotspots/analyze_hotspots.dart
```diff
@@ -0,0 +1,288 @@
...
+      body: SafeArea(
+        top: false,
+        left: false,
+        right: false,
+        child: Stack(
+          children: [
+            Column(
+              children: [
+                Expanded(
+                  child: ArcGISMapView(
```
PaulAllanSturm, line 86: should have the standard comment about ArcGISMapView above this line

lib/samples/analyze_hotspots/analyze_hotspots.dart
```diff
@@ -0,0 +1,288 @@
...
+                    controllerProvider: () => _mapViewController,
+                    onMapViewReady: onMapViewReady,
+                  ),
+                ),
+                Row(
+                  mainAxisAlignment: MainAxisAlignment.spaceEvenly,
+                  children: [
+                    ElevatedButton(
```
PaulAllanSturm: and comment above this line

lib/samples/analyze_hotspots/analyze_hotspots.dart
```diff
@@ -0,0 +1,288 @@
...
+                      onPressed: _analysisInProgress
+                          ? null
+                          : () => setState(() => _settingsVisible = true),
+                      child: const Text('Settings'),
+                    ),
+                    ElevatedButton(
```
PaulAllanSturm: and one here too

PaulAllanSturm (approved):

lib/samples/analyze_hotspots/analyze_hotspots.dart
jenmerritt, line 1: The UI feels a little clunky to me with having to tap down/back 2 levels each time.

What about:
- Having 2 buttons in the bottom row for "Select Date" and "Analyze".
- 'Select Date' pops open the date picker (does it need to be full screen?)
- 'Analyze re-runs the analysis.

Alternatively we could have a settings button only and then in the settings panel there is:
- 'Select Date'
- 'Analyze'

I'd be open to either, or other thoughts on this theme, but right now it doesn't quite feel right.

jenmerritt (commented):
The ArcGIS API is implemented nicely and good comments. 

I just have a query over the actual UI at the moment when I tried it out.

lib/samples/analyze_hotspots/analyze_hotspots.dart
martinoyovo: Thanks for the feedback, I updated the sample to use two bottom actions: "Select Date" and "Analyze", so there are no extra settings taps.

For the picker: on phones, Flutter’s `showDateRangePicker` is full screen by Material design, so that behavior is expected.

jenmerritt: Thanks @martinoyovo. Appreciate the comments and I think it's a lot cleaner now 👍  Understood about the picker.

lib/samples/analyze_hotspots/analyze_hotspots.dart
```diff
@@ -0,0 +1,268 @@
...
+                    ),
+                  ],
+                ),
+                Padding(
+                  padding: const EdgeInsets.only(top: 8, bottom: 12),
+                  child: Text(
+                    // Show the currently selected date range below the action buttons.
+                    'Selected date range: '
+                    '${formatDate(_selectedDateRange.start)} - '
+                    '${formatDate(_selectedDateRange.end)}',
```
jenmerritt:
```suggestion
                    '${formatDate(_selectedDateRange.start)} to '
                    '${formatDate(_selectedDateRange.end)}',
```

Nitpicking because it's a well written sample but I think using "to" here is easier to read since the dates are separated by `-`

jenmerritt (approved):
Approving - just one minor nitpick on the UI where the date is displayed, but trust you to make that change before merging!

martinoyovo:
Thanks @jenmerritt, I've updated the screenshot to reflect the change of "-"  character to "to". Merging now!

PR #510, opened 2026-05-20
Harish/add features with contingent values sample 5241

harish-kunchala:
### Add Features with ContingentValues Sample Implementation.


lib/samples/edit_feature_attachments/edit_feature_attachments.dart
jenmerritt, line 1: Just checking this is deliberately part of this PR

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      body: SafeArea(
+        top: false,
+        left: false,
+        right: false,
+        child: Stack(
+          children: [
+            ArcGISMapView(
```
jenmerritt, line 127:
```suggestion
            // Add a map view to the widget tree and set a controller.
            ArcGISMapView(
```
Standard sample comments

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+              controllerProvider: () => _mapViewController,
+              onMapViewReady: _onMapViewReady,
+              onTap: _onTap,
+            ),
+            LoadingIndicator(visible: !_ready),
```
jenmerritt, line 133:
```suggestion
            // Display a progress indicator and prevent interaction until state is ready.
            LoadingIndicator(visible: !_ready),
```
Standard sample comments

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+          ],
+        ),
+      ),
+      bottomSheet: _sheetVisible ? _buildBottomSheet(context) : null,
```
jenmerritt, line 138:
```suggestion
      // Show the settings as a bottom sheet when the map is tapped.
      bottomSheet: _sheetVisible ? _buildBottomSheet(context) : null,
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+
+  Future<void> _onMapViewReady() async {
+    // Set basemap from local VTPK.
+    final vtpkPath = _vtpkPath;
+    if (vtpkPath == null) return;
+
+    final vtpkLayer = ArcGISVectorTiledLayer.withUri(Uri.file(vtpkPath));
+    final basemap = Basemap.withBaseLayer(vtpkLayer);
+    final map = ArcGISMap.withBasemap(basemap);
+    _mapViewController.arcGISMap = map;
```
jenmerritt:
```suggestion
    // Create a vector tiled layer from a local file.
    final vtpkPath = _vtpkPath;
    if (vtpkPath == null) return;
    final vtpkLayer = ArcGISVectorTiledLayer.withUri(Uri.file(vtpkPath));
    
    // Create a basemap using the vector tiled layer and set to a map.
    final basemap = Basemap.withBaseLayer(vtpkLayer);
    final map = ArcGISMap.withBasemap(basemap);
    // Set the map to the controller.
    _mapViewController.arcGISMap = map;
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+
+    // Load the geodatabase + BirdNests table + contingent values definition.
+    await _loadGeodatabase();
```
jenmerritt:
```suggestion
    // Load the geodatabase and configure the table and contingent values definition.
    await _loadGeodatabase();
```
Perhaps amend this method to something like `_configureData`.

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+
+    // Add feature layer to map.
+    final featureLayer = FeatureLayer.withFeatureTable(_birdNestsTable);
+    await featureLayer.load();
+    _mapViewController.arcGISMap?.operationalLayers
+      ?..clear()
+      ..add(featureLayer);
```
jenmerritt: We don't need to get the map from the controller as we just created it as a variable. I also don't see why we need to clear the operational layers here as there isn't anywhere we could have added other layers. Plus we have the map as an object further up so we don't need to get it from the So I think you just need to call: `map.operationalLayers.add(featureLayer)`

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+    // Draw existing persisted buffer graphics.
+    await _refreshPersistedBuffers();
+
+    // Zoom to feature layer extent.
+    await _mapViewController.setViewpointGeometry(
+      featureLayer.fullExtent!,
+      paddingInDiPs: -20,
+    );
+
+    setState(() => _ready = true);
```
jenmerritt, line 182:
```suggestion
    // // Set the ready state variable to true to enable the sample UI.
    setState(() => _ready = true);
```
Standard sample comment

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+
+    _geodatabase = Geodatabase.withFileUri(File(gdbPath).uri);
+    await _geodatabase.load();
+
+    final table = _geodatabase.getGeodatabaseFeatureTable(
+      tableName: _tableName,
+    );
+    if (table == null) {
+      throw StateError('Could not find geodatabase table: $_tableName');
+    }
```
jenmerritt:
```suggestion
```

Not sure we want to be throwing if we're not catching this. The pattern in samples generally is that we're confident in the data enough that we can assume the table is not null. So I think just remove this.

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+  }
+
+  Future<void> _loadGeodatabase() async {
+    final gdbPath = _geodatabasePath;
+    if (gdbPath == null) return;
+
+    _geodatabase = Geodatabase.withFileUri(File(gdbPath).uri);
+    await _geodatabase.load();
+
+    final table = _geodatabase.getGeodatabaseFeatureTable(
```
jenmerritt: Can we just use the birdNestsTable here rather than reassigning lower down? i..e

`final _birdNestsTable = _geodatabase.getGeodatabaseFeatureTable(...`

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+      tableName: _tableName,
+    );
+    if (table == null) {
+      throw StateError('Could not find geodatabase table: $_tableName');
+    }
+
+    await table.load();
+
+    // Contingent values require the definition to be loaded.
+    await table.contingentValuesDefinition.load();
```
jenmerritt: This is something to test but jumps out - I'm curious if we need to load both the table and then the `contingentValuesDefinition` here. I'd suspect either loading the table does the latter for us, or we only need to load the `contingentValuesDefinition` but I might be wrong - so definitely test it out.

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+
+  // OnTap starts the draft edit session.
+  Future<void> _onTap(Offset offset) async {
+    if (!_ready) return;
+
+    // Convert screen point to map point.
+    final mapPoint = _mapViewController.screenToLocation(screen: offset);
+    if (mapPoint == null) return;
+
+    // Create a NEW draft feature (not inserted into the table yet).
```
jenmerritt:
```suggestion
    // Create a draft feature (not inserted into the table yet).
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+
+      _bufferMin = 0;
+      _bufferMax = 0;
+      _selectedBufferSize = null;
+
+      _isValid = false;
+      _sheetVisible = true;
+    });
+
+    // Show a draft point marker at the tap geometry.
```
jenmerritt:
```suggestion
    // Show a draft point marker at the tapped geometry.
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+  // Status options from coded value domain.
+  List<CodedValue> _statusOptions = const [];
+
+  // Currently selected status coded value.
+  CodedValue? _selectedStatus;
+
+  // Protection options (names) based on contingent values for the current draft.
+  List<String> _protectionOptions = const [];
+
+  // Currently selected protection name (user-facing).
```
jenmerritt:
```suggestion
  // Currently selected protection name.
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+      style: SimpleFillSymbolStyle.forwardDiagonal,
+      color: const Color(0xFFFF0000),
+      outline: outline,
+    );
+
+    _bufferOverlay.renderer = SimpleRenderer(symbol: fill);
+  }
+
+  void _configureDraftOverlay() {
+    // Draft point symbol (tap location)
```
jenmerritt:
```suggestion
    // Draft point symbol (tap location).
```
And multiple other places missing a period

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+
+    _draftFillSymbol = SimpleFillSymbol(
+      style: SimpleFillSymbolStyle.forwardDiagonal,
+      color: const Color(0xFFFF0000),
+      outline: outline,
+    );
+
+    // IMPORTANT:
+    // We use per-graphic symbols (point + polygon), so disable renderer
+    _draftOverlay.renderer = null;
```
jenmerritt: We aren't applying a renderer to this graphics overlay so I'm a little unclear a) why we need to set to null and b) why this is particularly important.

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,698 @@
...
+  // Draft graphics shown while user is editing a new feature.
+  final _draftOverlay = GraphicsOverlay();
+
+  // Draft point marker graphic for the tap location.
+  Graphic? _draftPointGraphic;
+
+  // Draft buffer graphic shown while the slider changes.
+  Graphic? _draftBufferGraphic;
+
+  // Draft feature (not added to table until Done).
```
jenmerritt:
```suggestion
  // Draft feature used when adding a new feature and is either discarded or added to the table while editing.
```

jenmerritt (commented):
Overall looks like it achieves the goal of the sample. 

The feedback is really on the polish and quality for presenting the sample - I'd encourage you to review the comments particularly, there's quite a lot that don't follow the pattern of our other samples which jump out a bit. I've picked out some specifics but as I say I encourage you to have a scan down as well.

lib/samples/edit_feature_attachments/edit_feature_attachments.dart
jenmerritt: Was it?

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,700 @@
+//
```
jenmerritt:
```suggestion
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,700 @@
...
+    // Draw existing persisted buffer graphics.
+    await _refreshPersistedBuffers();
+
+    // Zoom to feature layer extent.
+    await _mapViewController.setViewpointGeometry(
+      featureLayer.fullExtent!,
+      paddingInDiPs: -20,
+    );
+
+    // // Set the ready state variable to true to enable the sample UI.
```
jenmerritt:
```suggestion
    // Set the ready state variable to true to enable the sample UI.
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
jenmerritt: Was this not feasible? It's useful to get feedback / responses from you for when I come to a second review and then I understand if it's been missed or if there is a reason that it's stayed as it is - the latter is absolutely fine, it's meant to be a conversation.

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,700 @@
...
+    }
+  }
+
+  // Domain + contingent values
+
+  // Loads the Status field's coded value domain options.
+  void _loadStatusOptions() {
+    final statusField = _birdNestsTable.fields.firstWhere(
+      (f) => f.name == _fieldStatus,
+      orElse: () => throw Exception('Status field not found'),
```
jenmerritt: Is this exception being handled and displayed in the usual way we do in samples? I think there's a message dialog util.

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,700 @@
...
+
+                  setState(() {
+                    _selectedBufferSize = intVal;
+                    _isValid = false;
+                  });
+
+                  // Apply buffer size to draft attributes.
+                  feature.attributes[_fieldBufferSize] = intVal;
+
+                  // LIVE: update the draft buffer graphic as the slider changes.
```
jenmerritt:
```suggestion
                  // Update the draft buffer graphic as the slider changes.
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,700 @@
...
+                  // Apply protection code to draft attributes.
+                  _applyProtectionByName(name);
+
+                  // Refresh buffer range options.
+                  await _refreshBufferRange();
+                },
+        ),
+        const SizedBox(height: 12),
+
+        // Buffer size (contingent range values) + LIVE buffer preview.
```
jenmerritt:
```suggestion
        // Buffer size (contingent range values) and buffer preview.
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,700 @@
...
+                  // LIVE: update the draft buffer graphic as the slider changes.
+                  _updateDraftBufferGraphic(intVal.toDouble());
+
+                  // Validate contingent constraints.
+                  _validateDraft();
+                },
+        ),
+        const SizedBox(height: 12),
+
+        // Actions.
```
jenmerritt:
```suggestion
        // Buttons to either discard or save feature edits.
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,700 @@
...
+
+    // // Set the ready state variable to true to enable the sample UI.
+    setState(() => _ready = true);
+  }
+
+  Future<void> _configureData() async {
+    final gdbPath = _geodatabasePath;
+    if (gdbPath == null) return;
+
+    _geodatabase = Geodatabase.withFileUri(File(gdbPath).uri);
```
jenmerritt, line 190:
```suggestion
    // Create and load a geodatabase from a local file.
    _geodatabase = Geodatabase.withFileUri(File(gdbPath).uri);
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,700 @@
...
+  }
+
+  void _configureDraftOverlay() {
+    // Draft point symbol (tap location).
+    _draftPointSymbol = SimpleMarkerSymbol(
+      color: const Color(0xFF000000),
+      size: 11,
+    );
+
+    // Draft buffer symbol (same visual style as persisted buffers)
```
jenmerritt:
```suggestion
    // Draft buffer symbol (same visual style as persisted buffers).
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,700 @@
...
+      _draftOverlay.graphics.remove(_draftPointGraphic);
+      _draftPointGraphic = null;
+    }
+
+    final g = Graphic(geometry: geom, symbol: _draftPointSymbol);
+    _draftOverlay.graphics.add(g);
+    _draftPointGraphic = g;
+  }
+
+  void _clearDraftPointGraphic() {
```
jenmerritt:
```suggestion
  // Removes the graphic from the graphics overlay and resets the state.
  void _clearDraftPointGraphic() {
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,700 @@
...
+    if (_draftPointGraphic != null) {
+      _draftOverlay.graphics.remove(_draftPointGraphic);
+      _draftPointGraphic = null;
```
jenmerritt: Genuine question rather than a "do this" - is there any risk here to not having called `setState`? Or will where this is used always know the correct state of the `_draftPointGraphic` variable?

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,700 @@
...
+    }
+  }
+
+  void _clearDraftBufferGraphic() {
+    if (_draftBufferGraphic != null) {
+      _draftOverlay.graphics.remove(_draftBufferGraphic);
+      _draftBufferGraphic = null;
```
jenmerritt, line 303: Same question as above

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,700 @@
...
+
+  void _discardAndClose() {
+    // Draft-only: just clear graphics and reset state.
+    _endDraftEditing(dismissSheet: true);
+  }
+
+  // Ends draft editing by clearing draft graphics and resetting draft state.
+  void _endDraftEditing({required bool dismissSheet}) {
+    _clearDraftPointGraphic();
+    _clearDraftBufferGraphic();
```
jenmerritt: Since we call both of these methods at the same time, I wonder if we can just make one method called 

```
// Clears the graphics overlay and resets the draft graphic variables.
_clearGraphics
```

 and do:

```
_draftOverlay.graphics.clear().
_draftPointGraphic = null;
_draftBufferGraphic = null;
```

jenmerritt (commented):
The changes made look good, but there are a number of outstanding questions

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,704 @@
...
+      _AddFeaturesWithContingentValuesState();
+}
+
+class _AddFeaturesWithContingentValuesState
+    extends State<AddFeaturesWithContingentValues>
+    with SampleStateSupport {
+  // Controller for the map view.
+  final _mapViewController = ArcGISMapView.createController();
+
+  // A flag for when the local map view is ready and controls can be used.
```
jenmerritt:
```suggestion
  // A flag for when the map view is ready and controls can be used.
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
jenmerritt: One outstanding query here @harish-kunchala

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,704 @@
...
+  String? _selectedProtectionName;
+
+  // Buffer range derived from contingent range values.
+  int _bufferMin = 0;
+  int _bufferMax = 0;
+
+  // Currently selected buffer size.
+  int? _selectedBufferSize;
+
+  // Whether the current draft selections satisfy all contingencies.
```
jenmerritt:
```suggestion
  // Whether the current draft selections satisfy all contingencies and there are no violations.
```

jenmerritt: This just jumped out as "what does being valid mean here" - so reviewing the validation method I've made an additional suggestion below just to tie what this variable actually means to it being set - i.e. in relation to there not being violations. The extra attention to detail / nitpicking here is because this relates quite directly to the contingent values workflow this sample is working through.

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,704 @@
...
+    );
+    final ok = violations.isEmpty;
+
+    setState(() => _isValid = ok);
+
+    if (!ok && mounted) {
+      ScaffoldMessenger.of(context).showSnackBar(
+        const SnackBar(content: Text('Invalid contingent values')),
+      );
+    }
```
jenmerritt:
```suggestion
    // Capture whether the current draft selections satisfy all contingencies i.e. there are no violations
    setState(() => _isValid = violations.isEmpty);

    // Display a warning if the contingent valid are not valid.
    if (!_isValid && mounted) {
      ScaffoldMessenger.of(context).showSnackBar(
        const SnackBar(content: Text('Invalid contingent values')),
      );
    }
```
This just ties off my comment above about the meaning of the validation.

jenmerritt (commented):
Thanks @harish-kunchala , just a couple more thoughts

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,704 @@
...
+    final table = _geodatabase.getGeodatabaseFeatureTable(
+      tableName: _tableName,
+    );
+
+    if (table == null) return;
+
+    // Contingent values require the definition to be loaded.
+    await table.contingentValuesDefinition.load();
+
+    _birdNestsTable = table;
```
jenmerritt: My question above is outstanding but I'll try to explain via the code here. We don't need to add the null check as it's not a pattern we (currently) worry about with the samples as we are confident in the data, which doesn't change. So I think this can be simplified:

```
    _birdNestsTable = _geodatabase.getGeodatabaseFeatureTable(
      tableName: _tableName,
    )!;

    // Contingent values require the definition to be loaded.
    await _birdNestsTable.contingentValuesDefinition.load();
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.dart
```diff
@@ -0,0 +1,704 @@
...
+
+  // Validates contingent constraints for the current draft feature.
+  void _validateDraft() {
+    final feature = _draftFeature;
+    if (feature == null) return;
+
+    final violations = _birdNestsTable.validateContingencyConstraints(
+      feature: feature,
+    );
+    // Capture whether the current draft selections satisfy all contingencies i.e. there are no violations
```
jenmerritt:
```suggestion
    // Capture whether the current draft selections satisfy all contingencies i.e. there are no violations.
```

lib/samples/add_features_with_contingent_values/add_features_with_contingent_values.png
jenmerritt, line 1: The screengrab shouldn't have the header from the sample viewer - it should just be the sample content. E.g. https://github.com/Esri/arcgis-maps-sdk-flutter-samples/tree/main/lib/samples/display_map

jenmerritt (commented):
One conversation is resolved, 2 outstanding

jenmerritt (approved):
@changanxian this is ready for a second reviewer

changanxian (approved):

PR #512, opened 2026-05-27
Create Symbol Styles from Web Styles - update URL

jenmerritt:
- The doc URL was previously a 404. This PR updates it to the new reference. I've also created a PR to update the design where this was copied from.


jenmerritt:
Nudge

martinoyovo (approved):

PR #520, opened 2026-06-11
martinoyovo/clean up

martinoyovo:


harish-kunchala (approved):

jenmerritt:
I'd question this - what's the reason for combining the files? What's changed that means how it was before and previously reviewed is no longer correct?

PaulAllanSturm:
This sample is an outlier in that it is the only one that has multiple implementation files. That's not necessarily wrong, but it does add a weird extra wrinkle.

jenmerritt:
A weird extra wrinkle for what?  I just want to understand the motivation for the change other than noticing there's 2 files, as it wasn't an accident originally. If it's a) not adding clutter of irrelevant info to the sample file or b) making it harder for someone to directly copy the separate file then that's fine.

Not to make a big deal I just don't think there necessarily needs to be a rule that we have 1 file - unless there is some motivation for this.

jenmerritt:
Perhaps phrasing differently - do we see this as an outlier with our current snapshot of samples or as a pattern we want i.e. 1 file only? If the latter we should update our guidelines to be 1 file only - that way future samples follow this.

PaulAllanSturm:
I don't really have a strong opinion, but this is the only sample where you get "Prev." and "Next" buttons on the "Code" page. And the "import" line to bring in the second file is a little awkward, which means someone reading top-to-bottom might miss that chunk of code and have to spend a moment figuring out where it is. That's what I meant by extra wrinkle.
```
import 'package:arcgis_maps_sdk_flutter_samples/samples/show_device_location_with_nmea_data_sources/simulated_nmea_data_source.dart';
```

Screenshot 2026-06-12 at 9 32 42 AM

Also now I'm mad that we used "Prev." with a period instead of "Previous" on that page.

jenmerritt:
I don't either really, and have no objection to the change - it's more the approach that I'm attempting to encourage and question here. Thinking about why and the value - which ideally should be part of the PR description. Some justification along the lines of - it's tidier in this case because it's not immediately obvious to the user why this particular class would be in a separate file. But for other samples it may make more sense so we don't see the need for an additional guideline to restrict to 1 file (and our sample viewer can handle multiple files).

But at the moment it's like the PR has been made to clean up a mistake, whereas I think it was originally intentional. So I'm also considering the value of the original intent, I guess.

PaulAllanSturm (approved):

PR #530, opened 2026-07-07
Updated flutter_archive to current version.

harish-kunchala:
Updates the flutter_archive to the current version.

Built and ran our flutter samples app on both Android and iOS without any errors.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -8,8 +8,10 @@ environment:
   sdk: '>=3.12.1 <4.0.0'
 
 dependencies:
-  arcgis_maps: ^300.1.0
-  arcgis_maps_toolkit: ^300.1.0
+  arcgis_maps:
+    path: ../flutter/arcgis_maps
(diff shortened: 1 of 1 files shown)


martinoyovo (approved):

pubspec.yaml
jenmerritt, line 1: I'm seeing a warning that the dependencies aren't sorted alphabetically - since this PR is only amending pubspec.yaml and we're close to the release let's clean that up in this PR. Looks like app_settings should be at the top.

jenmerritt (approved):
Approving - if you're able to do the below clean up since we're close to release that would be great.

pubspec.yaml
harish-kunchala: Oh I agree, I wanted to discuss that with the team. If we should move app_settings to top.

jenmerritt: Thanks @harish-kunchala 👍

PR #533, opened 2026-07-14
Add a new sample: Display GeometryEditor Information During Interaction

changanxian:


lib/samples/display_geometry_editor_information_during_interaction/display_geometry_editor_information_during_interaction.dart
```diff
@@ -0,0 +1,457 @@
...
+      unawaited(subscription.cancel());
+    }
+
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      body: SafeArea(
```
PaulAllanSturm: Do we need a SafeArea for this sample? The only controls are in a floating `Positioned` widget.

lib/samples/display_geometry_editor_information_during_interaction/display_geometry_editor_information_during_interaction.dart
```diff
@@ -0,0 +1,457 @@
...
+                    _interactionPreviewValue != null)
+                  // Display information about the current interaction preview.
+                  Row(
+                    crossAxisAlignment: CrossAxisAlignment.start,
+                    children: [
+                      Text(
+                        _interactionPreviewDescription!,
+                        style: Theme.of(context).textTheme.labelMedium,
+                      ),
+                      const SizedBox(width: 8),
```
PaulAllanSturm: we should use the modern approach of `spacing: 8` as a `Row` property.

lib/samples/display_geometry_editor_information_during_interaction/display_geometry_editor_information_during_interaction.dart
changanxian: No. Removed

PaulAllanSturm (approved):

lib/samples/display_geometry_editor_information_during_interaction/display_geometry_editor_information_during_interaction.dart
```diff
@@ -0,0 +1,452 @@
...
+      _geometryEditor.onCanUndoChanged.listen(
+        (canUndo) => setState(() => _geometryEditorCanUndo = canUndo),
+      ),
+    );
+    _subscriptions.add(
+      _geometryEditor.onCanRedoChanged.listen(
+        (canRedo) => setState(() => _geometryEditorCanRedo = canRedo),
+      ),
+    );
+    _mapViewController.geometryEditor = _geometryEditor;
```
jenmerritt, line 145:
```suggestion
    // Set the geometry editor to the map view controller.
    _mapViewController.geometryEditor = _geometryEditor;
```

lib/samples/display_geometry_editor_information_during_interaction/display_geometry_editor_information_during_interaction.dart
```diff
@@ -0,0 +1,452 @@
...
+    // Ignore taps while editing is already in progress.
+    if (_geometryEditorIsStarted) return;
+
+    // Identify a graphic to edit at the tapped location.
+    final results = await _mapViewController.identifyGraphicsOverlay(
+      _graphicsOverlay,
+      screenPoint: localPosition,
+      tolerance: 12,
+    );
+    if (!mounted) return;
```
jenmerritt: Just curious about why this call is needed here? (It's probably correct, just for my education)

lib/samples/display_geometry_editor_information_during_interaction/display_geometry_editor_information_during_interaction.dart
```diff
@@ -0,0 +1,452 @@
...
+                  ],
+                ),
+                Row(
+                  spacing: 10,
+                  children: [
+                    Expanded(
+                      child: Tooltip(
+                        message: 'Undo',
+                        child: ElevatedButton(
+                          onPressed: _geometryEditorCanUndo ? onUndo : null,
```
jenmerritt: Could we just call `geometryEditor.undo()` here? Same for redo. Just seems like an unnecessary separate method, but happy to be disagreed with.

lib/samples/display_geometry_editor_information_during_interaction/display_geometry_editor_information_during_interaction.png
jenmerritt, line 1: Did you find the implementation of the controls in an "overlay" at the top was a good user experience? Vs our usual having them at the bottom below the map view? Again, more of a question to understand thought process. I know the design has them top right, but it's a desktop app.

I think we should also include the attribution in the screengrab per other samples.

lib/samples/display_geometry_editor_information_during_interaction/README.md
```diff
@@ -0,0 +1,50 @@
...
+6. Call `GeometryEditor.stop()` to finish the editing session and store the `Graphic`. The `GeometryEditor` does not automatically handle the visualization of a geometry output from an editing session. This must be done manually by propagating the geometry returned into a `Graphic` added to a `GraphicsOverlay`.
+    * To update the geometry underlying an existing `Graphic` in the `GraphicsOverlay`:
+        * Replace the existing `Graphic`'s `geometry` property with the geometry returned by the `GeometryEditor.stop()` method.
+
+## Relevant API
+
+* Geometry
+* GeometryEditor
+* GeometryEditor.onCanRedoChanged
+* GeometryEditor.onCanUndoChanged
```
jenmerritt: These aren't in the design

lib/samples/display_geometry_editor_information_during_interaction/display_geometry_editor_information_during_interaction.dart
changanxian: While that async identify operation is running, the user could leave the sample, the widget could be removed, or the map view could be disposed. When the await resumes, this State object may no longer be active.
The guard ensures that setState(...) updates occur on the correct widget.

lib/samples/display_geometry_editor_information_during_interaction/display_geometry_editor_information_during_interaction.dart
changanxian: yes

lib/samples/display_geometry_editor_information_during_interaction/display_geometry_editor_information_during_interaction.dart
jenmerritt: Does `SampleStateSupport` not guard against that? I mean I agree it's best practice to show in the code but I know that mixin was created to cover this scenario. I think it's one of the things we've talked about reviewing alongside more of the automation workflows - so probably a separate conversation.

changanxian: You are right, `setState` has had the guard. It doesn't hurt, but it's redundant; remove it.

lib/samples/display_geometry_editor_information_during_interaction/display_geometry_editor_information_during_interaction.png
changanxian: I prefer the current top-right overlay for this sample. These controls are tied directly to the geometry editing interaction, so keeping them over the map keeps the workflow immediate and avoids separating the live interaction feedback from the geometry being manipulated. Our usual bottom controls work well for sample settings, but here they would reduce map space and make the edit feedback feel less connected to the map interaction. Since the panel is compact and anchored away from the sample geometries, I think the overlay is a better UX for this case.

jenmerritt: Sounds good - thanks for the discussion! 👍

jenmerritt (approved):
LGTM

lib/samples/display_geometry_editor_information_during_interaction/display_geometry_editor_information_during_interaction.dart
PaulAllanSturm: A bit of discussion on SampleStateSupport in https://devtopia.esri.com/runtime/flutter/issues/5325 . This a really common problem so the Copilot reviewer is well-trained on it. Our too-clever "override setState in a mixin" approach is pretty obscure and the Copilot reviewer doesn't pick up on it.

PR #534, opened 2026-07-15
Monitor Changes to Map Load Status sample added.

kareemhassen:
New sample added that monitors a map and changes the load status dependent on the status of the map.


lib/samples/monitor_changes_to_map_load_status/monitor_changes_to_map_load_status.dart
```diff
@@ -0,0 +1,99 @@
...
+}
+
+class _MonitorChangesToMapLoadStatusState
+    extends State<MonitorChangesToMapLoadStatus>
+    with SampleStateSupport {
+  // Create a controller for the map view.
+  final _mapViewController = ArcGISMapView.createController();
+
+  // A string indicating the load status of the map.
+  String mapLoadStatus = 'Unknown';
```
jenmerritt:
```suggestion
  var mapLoadStatus = 'Unknown';
```
This isn't incorrect, but as a pattern we use `var` because Dart is smart enough to infer the type where we set an initial value. So here because we set a String, it knows.

lib/samples/monitor_changes_to_map_load_status/monitor_changes_to_map_load_status.dart
```diff
@@ -0,0 +1,99 @@
...
+
+    // Updates the map load status string based on the current LoadStatus and refreshes the UI.
+    updateMapStatus(LoadStatus.notLoaded);
+    try {
+      updateMapStatus(LoadStatus.loading);
+      await map.load();
+      updateMapStatus(LoadStatus.loaded);
+    } on Exception catch (_) {
+      updateMapStatus(LoadStatus.failedToLoad);
+    }
```
jenmerritt: When you set the map to the map view controller this actually starts to load the map.  There's some useful doc on loading patterns [here](https://developers.arcgis.com/flutter/programming-patterns/loading-resources-asynchronously/#loading).

So you want to wire up the UI updates _before_ you set the map.

That said, Copilot is right in stating that we should look to use `map.onLoadStatusChanged.listen(...)` here.

This [video](https://www.youtube.com/watch?v=nQBpOIHE4eE) gives some background on the concepts (and there is a whole playlist on async from flutterdev that might be worth watching. See if you can have a go at implementing.

Here's another [sample](https://github.com/Esri/arcgis-maps-sdk-flutter-samples/blob/main/lib/samples/add_raster_from_service/add_raster_from_service.dart) that uses a listener to give you an example. There are others too.

lib/samples/monitor_changes_to_map_load_status/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+* The map has a valid spatial reference.
+* The map has an an initial viewpoint.
+* One of the map's predefined layers has been created.
+
+## Relevant API
+
+* ArcGISMap
+* ArcGISMapView
+* LoadStatusChangedListener/LoadStatusChanged signal
```
jenmerritt: This is the part we aren't yet exercising, and again Copilot is right in stating we should update the relevant API to the Flutter specific calls - you can do this last once you've worked it into the code, as then you'll be clear on what API to include here.

lib/samples/monitor_changes_to_map_load_status/monitor_changes_to_map_load_status.dart
```diff
@@ -0,0 +1,109 @@
...
+  @override
+  void dispose() {
+    // Cancel the load status subscription when the widget is disposed.
+    _loadStatusSubscription?.cancel().ignore();
+    super.dispose();
+  }
+
+  // Updates the displayed map load status based on the current load state.
+  void updateMapStatus(LoadStatus status) {
+    if (!mounted) return;
```
jenmerritt: This isn't wrong, in fact it's very good practice. But we shouldn't need this line because we've got a `SampleStateSupport` mixin (see how  the sample `extends SampleStateSupport`). There are discussions about whether that's actually useful for developers that we made the decision to hide that away and it also is something AI struggles with but for now we have it.

Maybe one for @martinoyovo to explain further.

lib/samples/monitor_changes_to_map_load_status/monitor_changes_to_map_load_status.dart
```diff
@@ -0,0 +1,109 @@
...
+      };
+    });
+  }
+
+  Future<void> onMapViewReady() async {
+    // Create a map with an imagery basemap style
+    final map = ArcGISMap.withBasemapStyle(.arcGISImagery);
+
+    // Monitor changes to the map's load status.
+    _loadStatusSubscription?.cancel().ignore();
```
jenmerritt: Why are you canceling here? This should get cleaned up in the dispose. Was there an error or issue?

lib/samples/monitor_changes_to_map_load_status/monitor_changes_to_map_load_status.dart
```diff
@@ -0,0 +1,109 @@
...
+    // Set the map to the map view controller.
+    _mapViewController.arcGISMap = map;
+
+    // Trigger the load (the listener will update the UI as status changes).
+    try {
+      await map.load();
+    } on Exception catch (_) {
+      // Fallback in case an exception occurs before failed status is emitted.
+      updateMapStatus(LoadStatus.failedToLoad);
+    }
```
jenmerritt: `_mapViewController.arcGISMap = map;` -> this line / call actually triggers the map loading due to load dependencies in the `Loadable` model. So we don't need to then call `map.load()`.

We can remove the catch then, and instead, if the status is failed to load it will just get caught by the listener.

jenmerritt (commented):
Listener implementation looks good. There are just a couple of other refinements we can make.

lib/samples/monitor_changes_to_map_load_status/monitor_changes_to_map_load_status.dart
```diff
@@ -0,0 +1,102 @@
...
+    final map = ArcGISMap.withBasemapStyle(.arcGISImagery);
+
+    // Monitor changes to the map's load status.
+    _loadStatusSubscription = map.onLoadStatusChanged.listen(updateMapStatus);
+
+    // Set the map to the map view controller.
+    _mapViewController.arcGISMap = map;
+
+    // Load the map.
+    await map.load();
```
jenmerritt:
```suggestion
    // Set the map to the map view controller.
    // Note that this automatically loads the map as ArcGISMap.load is invoked by the map view controller in preparation for displaying the map.
    _mapViewController.arcGISMap = map;
```

Because `_mapViewController.arcGISMap = map` triggers loading I would actually amend as follows. This will help educate our developers on this pattern too. 

I actually think the readme could be improved to explain this better. I'm going to make a note to follow up on that in the design itself for another time.

lib/samples/monitor_changes_to_map_load_status/README.metadata.json
```diff
@@ -0,0 +1,28 @@
...
+        "ArcGISMap",
+        "ArcGISMap.onLoadStatusChanged",
+        "ArcGISMapView",
+        "LoadStatus"
+    ],
+    "snippets": [
+        "monitor_changes_to_map_load_status.dart"
+    ],
+    "title": "Monitor changes to map load status"
+}
```
jenmerritt:
```suggestion
}

```
The little icon in the git files indicates we should add a blank line at the end of this file.

jenmerritt (commented):
One final tweak then we're good.

For future PRs - what you can do once you've addressed changes and are ready for a re-review, click the little circular arrows next to the relevant reviewer. That then gives a fresh email notification "Kareem requested your review" and lets the reviewers know when you're done and to come back and check

Image

lib/samples/monitor_changes_to_map_load_status/monitor_changes_to_map_load_status.dart
martinoyovo: Thanks for the feedback! Kareem and I met yesterday and went through all of his PRs together, where I shared my comments and explained the changes needed. That’s why I didn’t add as many GitHub comments this time.

lib/samples/monitor_changes_to_map_load_status/monitor_changes_to_map_load_status.dart
kareemhassen: The reason I was canceling was I saw the AI recommended it and I was not too sure what the purpose was. Upon further investigation I realized the dispose already handled that case so I went ahead and got rid of that!

lib/samples/monitor_changes_to_map_load_status/monitor_changes_to_map_load_status.dart
kareemhassen: Will do! I will have the listener handle the status.

lib/samples/monitor_changes_to_map_load_status/monitor_changes_to_map_load_status.dart
kareemhassen: Got it! Yeah the reason why I have not been requesting a re-review was because I would push something and then after push I would realize that there is something else that I need to add. That is something bad on my end and I am sorry about that. Martino talked to me about it and I will do my best to make sure everything is 100% (up to my knowledge) and if not ask someone.

kareemhassen: In regard to the comment, " // Note that this automatically loads the map as ArcGISMap.load is invoked by the map view controller in preparation for displaying the map." 

I did not include this specific comment as I thought it was a comment for myself to further understand what is going on rather than something to include in the sample. If I understood that incorrectly, let me know and I will change it back.

martinoyovo (approved):

jenmerritt (approved):

lib/samples/monitor_changes_to_map_load_status/monitor_changes_to_map_load_status.dart
jenmerritt: Initially I was meaning for it to be there but in hindsight I think my comment about updating the design is more relevant here so I'm happy with where the sample is right now 👍

lib/samples/monitor_changes_to_map_load_status/monitor_changes_to_map_load_status.dart
jenmerritt: Yeah this is a great example of AI being too helpful and why it's important to understand why things are happening. If you asked it why it put it there it would probably reply "You're right! It's already being called in dispose" 🤦‍♀️

PR #538, opened 2026-07-20
New Sample: Add 3d Tiles Layer

kareemhassen:
Add a layer to visualize 3D tiles data that conforms to the OGC 3D Tiles specification.


lib/samples/add_3d_tiles_layer/add_3d_tiles_layer.dart
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  ArcGISScene _setupScene() {
+    // Create a scene.
+    final scene = ArcGISScene.withBasemapStyle(BasemapStyle.arcGISDarkGray);
+
+    // Set the scene's initial viewpoint.
+    scene.initialViewpoint = Viewpoint.withPointScaleCamera(
+      center: ArcGISPoint(x: 0, y: 0),
+      scale: 1,
```
jenmerritt, line 55: This jumps out as being quite generic values which makes me wonder if they're necessary - other than being required for this specific constructor. The design appears to just set a camera (using `setViewpointCamera`) and Swift uses the `Viewpoint.withExtentCamera` using the Camera's location as the bounding extent.

Curious if there was a reason you went with this constructor? One of those where multiple things achieve the same outcome.

jenmerritt (commented):
This looks good. I just have one query about how we're setting the viewpoint

lib/samples/add_3d_tiles_layer/add_3d_tiles_layer.dart
kareemhassen: The reason why I went with this constructor is because when it came time to implement the scene's initial viewpoint I went and looked at samples that were similar. One sample I found was `add_point_cloud_layer_from_file`. If you look at this sample, the way they implement the initial viewpoint they use the same values as me for the x and y and scale = 1, sort of like a "placeholder" as it is needed for this constructor. That is the reasoning behind it, if you'd like me to change it to be more compliant, I completely get that as well! I have attached a photo of the initialization that is present in the sample that I referenced. 
image

jenmerritt: I think it ultimately achieves the same so I'm ok with it - and it's not core to this sample. Given we have precedence elsewhere it would rather be something we looked at across the board, which is probably for minimal value. Thanks for your thoughtful response - this is exactly what these kind of PR discussions are for 👍

jenmerritt (approved):
Looks good to me!

martinoyovo (approved):

kareemhassen:
For the next commit I am doing it is just a small fix as I forgot to change `Future<void> onMapViewReady() async` to `void onMapViewReady()` since there is no await present inside the void.

lib/samples/add_3d_tiles_layer/README.metadata.json
```diff
@@ -0,0 +1,27 @@
...
+    "redirect_from": [],
+    "relevant_apis": [
+        "ArcGISSceneView",
+        "Ogc3DTilesLayer"
+    ],
+    "snippets": [
+        "add_3d_tiles_layer.dart"
+    ],
+    "title": "Add 3D tiles layer"
+}
```
martinoyovo: Missing an additional line at the end of this file.

kareemhassen: Will fix right now.

lib/samples/add_3d_tiles_layer/add_3d_tiles_layer.dart
```diff
@@ -0,0 +1,84 @@
...
+    );
+  }
+
+  void onSceneViewReady() {
+    // Add the scene to the view controller.
+    _sceneViewController.arcGISScene = _setupScene();
+  }
+
+  ArcGISScene _setupScene() {
+    // Create a scene.
```
martinoyovo:
```suggestion
    // Create a Scene with the ArcGIS DarkGrayBase basemap style.
```

martinoyovo (changes requested):

lib/samples/add_3d_tiles_layer/add_3d_tiles_layer.dart
```diff
@@ -0,0 +1,84 @@
...
+    );
+  }
+
+  void onSceneViewReady() {
+    // Add the scene to the view controller.
+    _sceneViewController.arcGISScene = _setupScene();
+  }
+
+  ArcGISScene _setupScene() {
+    // Create a Scene with the ArcGIS DarkGrayBase basemap style.
```
martinoyovo:
```suggestion
    // Create a Scene with the ArcGIS DarkGray basemap style.
```

martinoyovo: Just a small tweak here

martinoyovo (changes requested):

martinoyovo (approved):
LGTM to me

kareemhassen:
Woo! Finally!

jenmerritt (approved):
Nice, and good catch on the `onMapViewReady` clean up 👍 

Onto the next!!! 🎉

PR #541, opened 2026-07-22
Support Android 17

martinoyovo:
### Summary

Bumps the Android compile SDK to 37 so the Flutter samples app supports Android 17.

### DIFF
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -5,7 +5,7 @@ plugins {
 
 android {
     namespace = "com.esri.arcgis_maps_sdk_flutter_samples"
-    compileSdk = 36
+    compileSdk = 37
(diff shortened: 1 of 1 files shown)


martinoyovo:
I've done a final round of testing for the Android 17 update.

When I bumped the Flutter repo's compileSdk to 37, the Toolkit Example app ran successfully on Android 17. I then tested the Sample Viewer, ArcGIS Maps example app, Sandbox, and Internal apps, and they also ran successfully.

After that, I reverted the Flutter repo back to compileSdk 36 and kept the Sample Viewer, Toolkit, and other repos on compileSdk 36 as well. I was still able to run all of the apps successfully on Android 17.

Based on this testing, our SDK works correctly on Android 17, and keeping our compileSdk version at 36 appears to be okay for now

PR #544, opened 2026-07-24
Refactor all banners to a single theme-aware banner widget

PaulAllanSturm:
- Introduces a new widget, `MapBanner`, which displays a string of text on a semi-transparent background that stretches the full width, and is theme-aware. Several samples were doing this, with slight variations, and some with bad interactions with Dark Mode. They have all now been standardized on `MapBanner`.
- Removed some text styles that were no longer in use and had hard-coded white rather than something theme-aware.
- Removed a trailing newline in "Navigate route" sample's banner message, which was adding unnecessary padding on the bottom.
- Removed a couple `SafeArea`s from samples that didn't have any widgets that would make a `SafeArea` necessary.

### DIFF
--- a/lib/common/common.dart
+++ b/lib/common/common.dart
@@ -20,6 +20,7 @@ export 'dialogs.dart';
 export 'download_util.dart';
 export 'feature_popup_view.dart';
 export 'loading_indicator.dart';
+export 'map_banner.dart';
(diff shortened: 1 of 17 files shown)


martinoyovo (approved):
Looks great!

changanxian (approved):

PR #546, opened 2026-07-27
Sample: Add vector tiled layer from custom style

PaulAllanSturm:
Implements the "Add vector tiled layer from custom style" sample. Used Copilot to create a draft (it came out fully functional) and then sanded it down a little.


lib/samples/add_vector_tiled_layer_from_custom_style/add_vector_tiled_layer_from_custom_style.dart
```diff
@@ -0,0 +1,268 @@
...
+              children: [
+                Expanded(
+                  // Add a map view to the widget tree and set a controller.
+                  child: ArcGISMapView(
+                    controllerProvider: () => _mapViewController,
+                    onMapViewReady: onMapViewReady,
+                  ),
+                ),
+                // Add a style picker to switch between online and offline styles.
+                Row(
```
martinoyovo: Since the `Row` only contains a single child, it doesn't appear to be providing much value here. We could replace it with `Center` or remove it entirely if the parent already handles alignment.

PaulAllanSturm: fixed

martinoyovo (approved):

jenmerritt (approved):

PR #561, opened 2026-08-10
New Sample Added: ConfigureBasemapStyleParameters

kareemhassen:
New sample added that applies basemap style parameters customization for a basemap, such as displaying all labels in a specific language or displaying every label in their corresponding local language.


lib/samples/configure_basemap_style_parameters/configure_basemap_style_parameters.dart
```diff
@@ -0,0 +1,250 @@
...
+    // Create and apply a new basemap using the configured parameters.
+    _mapViewController.arcGISMap?.basemap = Basemap.withStyle(
+      BasemapStyle.arcGISLightGray,
+      parameters: parameters,
+    );
+  }
+
+  // Displays a bottom sheet for selecting a language strategy or a specific language.
+  Future<void> _showLanguageSettings() async {
+    await showModalBottomSheet<void>(
```
jenmerritt: A question rather than suggestion - do we want the modal bottom sheet here, as opposed to the bottom sheet util we have within the samples?

lib/samples/configure_basemap_style_parameters/README.md
```diff
@@ -0,0 +1,46 @@
+# Configure basemap style parameters
+
+Apply basemap style parameters customization for a basemap, such as displaying all labels in a specific language or displaying every label in their corresponding local language.
+
+![Configure basemap style parameters](configure_basemap_style_parameters.png)
```
jenmerritt:
```suggestion
![Image of configure basemap style parameters](configure_basemap_style_parameters.png)
```

lib/samples/configure_basemap_style_parameters/configure_basemap_style_parameters.dart
PaulAllanSturm: The big difference with our `BottomSheetSettings` widget is that is non-modal. Which is to say, you can still interact with the map behind it while playing around with the settings. On the other hand, `showModalBottomSheet` precludes you from interacting with the map until the sheet is dismissed. We almost always want the non-modal sheet so that you can examine the outcome as you explore the settings. We would only want a modal sheet if there's a technical reason you need to complete the task in the sheet and only apply the changes to the map when the sheet gets dismissed.

kareemhassen: Got it, responding to Jen, I did not know that `BottomSheetSettings` was a thing and in practice it makes more sense because it allows the user to manipulate values without having to reopen the BottomSheet every time. The only reason I used modal bottom sheet is that is what I used in another sample but I completely agree that `BottomSheetSettings` makes more sense in this scenario.

lib/samples/configure_basemap_style_parameters/configure_basemap_style_parameters.dart
```diff
@@ -0,0 +1,250 @@
...
+class ConfigureBasemapStyleParameters extends StatefulWidget {
+  const ConfigureBasemapStyleParameters({super.key});
+
+  @override
+  State<ConfigureBasemapStyleParameters> createState() =>
+      _ConfigureBasemapStyleParametersState();
+}
+
+// Represents a basemap language strategy option displayed in the UI.
+class Strategy {
```
martinoyovo: We should make the `Strategy` and `Language` classes private to avoid any future samples calling them by error (with an underscore in front. ie: `_Strategy` and `_Language`).

Edit: Copilot suggested: `_LanguageStrategyOption` and `_SpecificLanguageOption`, that's even better

lib/samples/configure_basemap_style_parameters/configure_basemap_style_parameters.dart
```diff
@@ -0,0 +1,250 @@
...
+                          // Update the selected strategy and clear any specific language.
+                          _selectedLanguageStrategy =
+                              languageStrategy.languageStrategy;
+                          _selectedSpecificLanguage = '';
+                        });
+                        // Apply the updated language setting.
+                        _setBasemapLanguage();
+                        Navigator.pop(context);
+                      },
+                    );
```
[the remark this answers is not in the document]
martinoyovo: I think Copilot feedback here is valuable and we should implement that. We've talked about the `RadioListTile` specific widget but we didn't want to implement it until we were sure the logic behind the code was right. Now that Copilot suggested it, it's appropriate we follow that.

lib/samples/configure_basemap_style_parameters/configure_basemap_style_parameters.dart
kareemhassen: Got it! Implemented that into the next commit.

lib/samples/configure_basemap_style_parameters/configure_basemap_style_parameters.dart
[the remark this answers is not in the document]
kareemhassen: When implementing I get `'groupValue' is deprecated and shouldn't be used. Use a RadioGroup ancestor to manage group value instead. This feature was deprecated after v3.32.0-0.0.pre.
Try replacing the use of the deprecated member with the replacement.` which means we do not use `RadioListTile` anymore.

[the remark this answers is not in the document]
martinoyovo: Makes sense. In our case, we might need to replace the `ExpansionTile` with a `RadioGroup`. That'd serve as a parent here and its children could now be `RadioListTile`s.

jenmerritt (approved):

lib/samples/configure_basemap_style_parameters/configure_basemap_style_parameters.dart
```diff
@@ -0,0 +1,250 @@
...
+              children: [
+                Expanded(
+                  // Add a map view to the widget tree and set a controller.
+                  child: ArcGISMapView(
+                    controllerProvider: () => _mapViewController,
+                    onMapViewReady: onMapViewReady,
+                  ),
+                ),
+                // Add a language settings button to the widget tree.
+                ElevatedButton(
```
martinoyovo, line 123: We should wrap this button in a Center widget. The alignment may look okay on iOS and may not on Android

kareemhassen: Fixed.

martinoyovo (approved):

PR #562, opened 2026-08-11
Introduce new layout for storing downloadable resources on disk

PaulAllanSturm:
A new approach to downloading offline resources.
- Each downloadable resource has a `PortalItem` that we can leverage. So we load the `PortalItem` and cache it to disk in JSON form.
- The `PortalItem.name` property is the filename of the attached data file. We had been manually providing that as the "downloadable" value in the `README.metadata.json` files, but we can just use `PortalItem.name` instead.
- Now that the "downloadable" value in the JSON is no longer necessary, it is removed from all the `README.metadata.json` files that had it. (That's why there's so many files changed in this PR -- those are just the one line changes to remove "downloadable".)
- Files are now stored in a new layout on disk. All files are placed under the "OfflineDataLocation" directory. Inside there, the PortalItem json is stored as "`${portalItem.itemId}.json`". The downloaded files are in the "`${portalItem.itemId}.downloaded/`" directory. This approach lets us associate the PortalItem with the downloaded files by using the unique `itemId` value. That'll come in handy later when we provide a UI for managing the downloaded files (i.e., we want to be able to delete all downloaded data or delete/re-download resources associated with a particular sample).


changanxian (approved):
LGTM

kmueller-gis (approved):

PR #572, opened 2026-08-27
"Create and save map" sample

PaulAllanSturm:
Implements the "Create and save map" sample.


lib/samples/create_and_save_map/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+![Image of create and save map](create_and_save_map.png)
+
+## Use case
+
+Maps can be created programmatically in code and then serialized and saved as an ArcGIS portal item. In this case, the portal item is a web map which can be shared with others and opened in various applications and APIs throughout the platform, such as ArcGIS Pro, ArcGIS Online, the JavaScript API, Collector, and Explorer.
+
+## How to use the sample
+
+When you run the sample, you will be challenged for an ArcGIS Online login. Enter a username and password for an ArcGIS Online named user account (such as your ArcGIS for Developers account). Then, choose the basemap and layers for your new map. To save the map, add a title, tags and description (optional), and a folder on your portal (you will need to create one in your portal's My Content section if you don't already have one). Click the Save to Portal button to save the map to the chosen folder.
```
jenmerritt:
```suggestion
When you run the sample, you will be challenged for an ArcGIS Online login. Enter a username and password for an ArcGIS Online named user account (such as your ArcGIS for Developers account). Then, choose the basemap and layers for your new map. To save the map, add a title, tags and description (optional), and a folder on your portal (you will need to create one in your portal's My Content section if you don't already have one). Tap the "Save to Portal" button to save the map to the chosen folder.
```
That old chestnut

lib/samples/create_and_save_map/create_and_save_map.dart
```diff
@@ -0,0 +1,317 @@
...
+  State<CreateAndSaveMap> createState() => _CreateAndSaveMapState();
+}
+
+// A convenience class to record the selection state of a map layer.
+class LayerRecord {
+  LayerRecord(String uri) : layer = ArcGISMapImageLayer.withUri(Uri.parse(uri));
+
+  final ArcGISMapImageLayer layer;
+  bool selected = false;
+}
```
jenmerritt, line 29: Is it normal enough to put a class between the widget and state? Probably fine but jumped out (could just be a Friday afternoon thing)

lib/samples/create_and_save_map/create_and_save_map.dart
```diff
@@ -0,0 +1,317 @@
...
+  // Create an OAuthUserConfiguration.
+  // This document describes the steps to configure OAuth for your app:
+  // https://developers.arcgis.com/documentation/security-and-authentication/user-authentication/flows/authorization-code-with-pkce/
+  final _oauthUserConfiguration = OAuthUserConfiguration(
+    portalUri: Uri.parse('https://www.arcgis.com'),
+    clientId: 'T0A3SudETrIQndd2',
+    redirectUri: Uri.parse('my-ags-flutter-app://auth'),
+  );
+  // Set aside the API Key so that it does not interfere with portal authentication.
+  late String _rememberedApiKey;
```
jenmerritt:
```suggestion
  // Set aside the sample viewer application API Key so that it does not interfere with portal authentication used within this sample.
  late String _sampleViewerApiKey;
```

I take it this is a sample viewer specific workflow? I think we should be a bit more verbose here as it took me a while to figure out. And maybe move this property right down to the bottom (rather than in amongst the sample specific properties)

lib/samples/create_and_save_map/create_and_save_map.dart
```diff
@@ -0,0 +1,317 @@
...
+  // A flag to indicate whether the map view is ready.
+  var _ready = false;
+
+  @override
+  void initState() {
+    super.initState();
+
+    // Set aside the API Key so that portal authentication is used.
+    _rememberedApiKey = ArcGISEnvironment.apiKey;
+    ArcGISEnvironment.apiKey = '';
```
jenmerritt:
```suggestion
    // Temporarily unset the sample viewer application API Key so that portal authentication is used for this sample. Store it to reset on dispose.
    _rememberedApiKey = ArcGISEnvironment.apiKey;
    ArcGISEnvironment.apiKey = '';
```
(and property rename per above).

lib/samples/create_and_save_map/create_and_save_map.dart
```diff
@@ -0,0 +1,317 @@
...
+    Authenticator.revokeOAuthTokens()
+        .catchError((Object error) {
+          // This sample has been disposed, so we can only report errors to the console.
+          // ignore: avoid_print
+          print('Error revoking tokens: $error');
+        })
+        .whenComplete(Authenticator.clearCredentials)
+        .ignore();
+
+    // Restore the API Key.
```
jenmerritt:
```suggestion
    // Restore the sample viewer application API Key.
```

lib/samples/create_and_save_map/README.md
PaulAllanSturm: derp fixed

lib/samples/create_and_save_map/create_and_save_map.dart
PaulAllanSturm: Normally we put these helper types at the bottom, but sometimes at the top. Never in between, though. I'll move this one to the top -- that makes it cleaner when you're reading the sample from top-to-bottom, so you don't run into this type and think it's part of the API or have to wonder what it is.

lib/samples/create_and_save_map/create_and_save_map.dart
PaulAllanSturm: fixed

lib/samples/create_and_save_map/create_and_save_map.dart
jenmerritt: Sounds good thanks!

jenmerritt (approved):

martinoyovo (approved):

PR #575, opened 2026-09-02
Harish/identify kml features 5701

harish-kunchala:
- Created Identify KML Features Sample
- There are no significant weather updates for Sep 2 2026, hence the screenshot describes that.


AGENT_REPORT.md
jenmerritt, line 1: Should this be git ignored / removed from the PR?

lib/samples/identify_kml_features/README.md
```diff
@@ -0,0 +1,49 @@
...
+## Relevant API
+
+* ArcGISMapView
+* ArcGISMapViewController.callout
+* ArcGISMapViewController.identifyLayer
+* IdentifyLayerResult
+* KmlDataset
+* KmlLayer
+* KmlPlacemark
+* KmlPlacemark.balloonContent
```
jenmerritt: This doesn't match the design

lib/samples/identify_kml_features/identify_kml_features.png
jenmerritt, line 1: Are there no severe weather alerts anywhere in the US even if you zoom / pan around a bit? I guess ideally we'd show the layer and callout but if there isn't any weather at all then not much we can do. I wondered if there might be a late summer thunderstorm somewhere! Probably no freezing rain this time of year... ⛄

lib/samples/identify_kml_features/identify_kml_features.dart
```diff
@@ -0,0 +1,180 @@
...
+            onTap: _onTap,
+          ),
+          // Prompt the user to identify a KML feature.
+          const SafeArea(
+            child: Align(
+              alignment: Alignment.topCenter,
+              child: Card(
+                child: Padding(
+                  padding: EdgeInsets.all(8),
+                  child: Text('Tap a weather feature to view its details.'),
```
jenmerritt: Maybe we can use the `MapBanner` util for this message?

lib/samples/identify_kml_features/identify_kml_features.dart
```diff
@@ -0,0 +1,180 @@
...
+      ),
+    ),
+  );
+
+  // A flag for when the map view is ready and controls can be used.
+  var _ready = false;
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
```
jenmerritt, line 47: Are we missing a SafeArea?

lib/samples/identify_kml_features/identify_kml_features.dart
```diff
@@ -0,0 +1,180 @@
...
+  const IdentifyKmlFeatures({super.key});
+
+  @override
+  State<IdentifyKmlFeatures> createState() => _IdentifyKmlFeaturesState();
+}
+
+class _IdentifyKmlFeaturesState extends State<IdentifyKmlFeatures>
+    with SampleStateSupport {
+  // Create a map with a dark gray basemap style.
+  final _map = ArcGISMap.withBasemapStyle(.arcGISDarkGrayBase);
```
jenmerritt: Are you needing access to the map anywhere other than in `onMapViewReady`? If not I think we can just move the map instantiation there.

lib/samples/identify_kml_features/README.md
harish-kunchala: Updated API according to the design.

lib/samples/identify_kml_features/identify_kml_features.png
harish-kunchala: I can change the time temporarily and do that, if that works. But for the current one. There doesn't seem to be anything.

jenmerritt: Yeah great thinking, worth a try (could set it to the date of the swift one where there was some weather!)

jenmerritt (commented):
Changes made look good, will await reviewing the other comments that haven't yet been addressed

lib/samples/identify_kml_features/identify_kml_features.dart
harish-kunchala: Done.

lib/samples/identify_kml_features/identify_kml_features.dart
harish-kunchala: No the Map can use the fullScreen. so not a problem.

lib/samples/identify_kml_features/identify_kml_features.dart
harish-kunchala: Done.

lib/samples/identify_kml_features/identify_kml_features.png
harish-kunchala: There's a problem with the changing the URL to get the data. 
when I use the current date URL: https://www.wpc.ncep.noaa.gov/kml/noaa_chart/WPC_Day1_SigWx_latest.kml

### Current Date:
I get the NOAA and NWS logos. 
identify-cur-date

### Different Date:
**But as soon as I change this to different date. We lose the logos.**
identify-diff-date

jenmerritt: Ok strange never mind, we'll just have to go with no weather data - to clarify, I mean the initial screengrab where there is no weather but the screen overlays are present. The screengrab needs updating to reflect the updated banner text however.

harish-kunchala: Great. I've updated the image now.

lib/samples/identify_kml_features/identify_kml_features.dart
```diff
@@ -0,0 +1,184 @@
...
+          fullExtent,
+          paddingInDiPs: 25,
+        );
+      }
+    } on Exception catch (exception) {
+      // Report a KML layer loading failure to the user.
+      showExceptionDialog('Failed to load KML layer', exception);
+    } finally {
+      // Hide the loading indicator after the layer finishes loading.
+      if (mounted) setState(() => _ready = true);
```
martinoyovo: I don't think we need the `mounted` check here. `SampleStateSupport` overrides `setState` and checks mounted internally, so the explicit guard is redundant

lib/samples/identify_kml_features/identify_kml_features.dart
```diff
@@ -0,0 +1,184 @@
...
+    }
+  }
+
+  Future<void> _onTap(Offset screenPoint) async {
+    if (!_ready) return;
```
martinoyovo, line 93: `LoadingIndicator` expands over the stack and its ColoredBox intercepts taps while visible, so `_onTap` cannot run before readiness through this UI. We could remove the redundant guard

harish-kunchala: I agree redundant through the current UI, but I’d keep it so `_onTap` enforces its own readiness precondition rather than relying on the loading overlay’s hit-testing behavior.

lib/samples/identify_kml_features/identify_kml_features.dart
harish-kunchala: I agree.  removed the `mounted` check.

martinoyovo (approved):

jenmerritt (approved):

PR #579, opened 2026-09-09
"Add KML layer with network links" sample

PaulAllanSturm:
Implements the "Add KML layer with network links" sample.


lib/samples/add_kml_layer_with_network_links/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Relevant API
+
+* KmlDataset
+* KmlDataset.onNetworkLinkMessageReceived
+* KmlLayer
+
+## Offline data
+
+This sample uses the radar.kmz file, which can be found on [ArcGIS Online](https://arcgisruntime.maps.arcgis.com/home/item.html?id=600748d4464442288f6db8a4ba27dc95).
```
martinoyovo: We mention the offline data here but we didn't add it to the metadata.json. Should we route this through the resource download mechanism we already have?

PaulAllanSturm: Hmm that's a good question. I checked the `implementation-details.md`, the Swift version, and the Kotlin version. They all use the `KmlDataset` constructor that takes a network URL -- none of them download the file. Plus the file is mostly just a pointer to a live service to get live aviation data. If anything, we should get rid of the "Offline data" section and add that little bit to the "About the data" section. Let's see what Jen says when she looks at this.

martinoyovo: Agreed. The sample doesn't actually download or use offline data, so removing that section and adding a note under "About the data" seems more accurate. Let's see what Jen says.

jenmerritt (approved):

lib/samples/add_kml_layer_with_network_links/README.md
jenmerritt: The design itself (.NET) in the implementation details as you mention uses the URL constructor so yeah I'd just move that line into "About the data".

@martinoyovo do you want to take on creating a small PR in common-samples to do the same in the design? So it becomes:

```
About the data

This map shows the current air traffic in parts of Europe with heading, altitude, and ground speed. Additionally, noise levels from ground monitoring stations are shown.

This sample uses the radar.kmz file, which can be found on [ArcGIS Online](https://arcgisruntime.maps.arcgis.com/home/item.html?id=600748d4464442288f6db8a4ba27dc95).
```

martinoyovo: That's a great idea @jenmerritt, I'll do that.

martinoyovo (approved):

PR #585, opened 2026-09-18
Open KML Node list to show selected node

kmueller-gis:
Based on tester feedback, the KML Node List sample has been updated to open the KML node list to the currently selected node.
- List scrolls to the lowest level parent list tile, but not to the tile for the selected node.


lib/samples/list_contents_of_kml_file/list_contents_of_kml_file.dart
```diff
@@ -161,6 +165,7 @@ class _ListContentsOfKmlFileState extends State<ListContentsOfKmlFile>
           child: Text(kmlNode.name),
         ),
         subtitle: Text(_getFriendlyTypeName(kmlNode)),
+        initiallyExpanded: _selectedNodeAncestors.contains(kmlNode),
```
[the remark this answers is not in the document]
kmueller-gis: There is no straight forward way to do this. Adding this extra UI logic goes beyond the point of the sample which is illustrate how to list the nodes in a KML file.

lib/samples/list_contents_of_kml_file/list_contents_of_kml_file.dart
```diff
@@ -318,4 +326,21 @@ class _ListContentsOfKmlFileState extends State<ListContentsOfKmlFile>
       return Viewpoint.fromTargetExtent(extentBuilder.extent);
     }
   }
+
+  // Recursive function to build list of ancestors for a given KML node. List
+  // includes the initial node.
+  List<KmlNode> _buildAncestorNodeList(KmlNode kmlNode) {
```
PaulAllanSturm, line 332: Er well walking *up* a tree doesn't require recursion
```dart
final ancestors = <KmlNode>[];
KmlNode? currentNode = kmlNode;
while (currentNode != null) {
  ancestors.add(currentNode);
  currentNode = currentNode.parent;
}
return ancestors;
```
There's probably even some clever one-liner using some clever map/reduce chain but I dunno, recursion is overkill here

PaulAllanSturm: lol how about
```dart
Iterable<KmlNode> walkUp(KmlNode? node) sync* {
  if (node == null) return;
  yield node;
  yield* walkUp(node.parent);
}
return walkUp(kmlNode).toList();
```

kmueller-gis: Replaced recursion with a while loop.

PaulAllanSturm (approved):

martinoyovo (approved):

